PR #141, opened 2024-09-03
Fix `code` tag growing layout

dan-lee:
The `grid` is a leftover when the header still was `position: fixed` and can be removed. For some reason, this made the `code` tags in MDX grow uncontrollably

### DIFF
--- a/packages/zudoku/src/app/main.css
+++ b/packages/zudoku/src/app/main.css
@@ -96,7 +96,7 @@
   }
 
   #root {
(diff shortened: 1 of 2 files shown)


packages/zudoku/src/app/main.css
```diff
@@ -96,7 +96,7 @@
   }
 
   #root {
-    @apply min-h-screen grid grid-rows-[var(--header-height)_1fr] w-full;
```
mosch, line 99: 🤦

dan-lee: 😢

PR #142, opened 2024-09-03
Update the docs for GitHub Pages

vazexqi:
Use the `basePath` feature to allow publishing on GitHub Actions.


package-lock.json
dan-lee, line 1: We are using `pnpm` in Zudoku :) You can remove this file

vazexqi: Removed in the latest commit

PR #186, opened 2024-09-06
[feat] Added sitemaps to docs + enhanced sitemap config

ntotten:
- Added sitemaps to website and docs
- Added additional configuration options for sitemaps
- Documented sitemap configuration: https://zudoku.dev/docs/configuration/overview

### DIFF
--- a/packages/zudoku/src/vite/prerender.ts
+++ b/packages/zudoku/src/vite/prerender.ts
@@ -104,7 +104,7 @@ export const prerender = async (html: string, dir: string) => {
     const toUrlLoc = (path?: string) => {
       try {
         if (!path || path.includes("*")) return [];
-        return `<url><loc>${new URL(joinPath(config.basePath, path), config.sitemap!.baseUrl)}</loc></url>`;
+        return `<url><loc>${new URL(joinPath(config.basePath, path), config.sitemap!.siteUrl)}</loc></url>`;
       } catch {
         // skip invalid urls
         return [];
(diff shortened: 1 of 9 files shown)


dan-lee (approved):

packages/zudoku/src/vite/prerender.ts
```diff
@@ -100,34 +101,12 @@ export const prerender = async (html: string, dir: string) => {
   // eslint-disable-next-line no-console
   console.log(`Prerendered ${paths.length} pages`);
 
-  if (config.sitemap) {
-    const toUrlLoc = (path?: string) => {
-      try {
-        if (!path || path.includes("*")) return [];
```
dan-lee, line 106: @ntotten Are catch-all paths still skipped?

packages/zudoku/src/vite/sitemap.ts
```diff
@@ -0,0 +1,74 @@
...
+      ? await config.exclude()
+      : config.exclude) ?? [];
+
+  for (const url of outputUrls) {
+    if (!exclude.includes(url) && !url.includes("*")) {
+      sitemap.write({
+        url: new URL(joinPath(basePath, url), config.siteUrl).toString(),
+        lastmod,
+        changefreq: config.changefreq ?? "daily",
+        priority: config.priority ?? 0.7,
```
dan-lee, line 65: Why do we default here instead of leaving it blank? Google ignores these anyway: https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap

ntotten: Because this is what everyone else seems to do. :) I think it will make people feel better even if google doesnt care. next-sitemap, etc. all fill this stuff in. IMO it makes the sitemap feel a bit more complete even if its pointless for google.

PR #187, opened 2024-09-06
Use `tsx` to load zudoku config

dan-lee:
While developing locally with importing TSX files in the `zudoku.config.tsx`, we had found some issues:

Switching to use [`tsImport`](https://tsx.is/dev-api/ts-import) from `tsx` (which we use anyway) got us away from the hack-ish approach loading the config with Vite's `loadConfigFromFile` and solved the issue.


mosch:
Lgtm, @ntotten  any thoughts?

mosch (approved):

PR #193, opened 2024-09-06
[fix] Fixes CLI to correctly check for outdated versions

ntotten:


lcampos (commented):
Reviewed and approved

PR #194, opened 2024-09-06
[fix] Fixes create-zudokou-app to use correct version

ntotten:


lcampos (commented):
Reviewed and approved

PR #196, opened 2024-09-06
Read package.json from file in create-zudoku-app

ntotten:
Reads the create-zudoku-app package.json. I dont like the bundling because it makes the build vs version order too dependant. This keeps it so that no matter when the package.json gets updated the app will get the correct version.


lcampos (commented):
Reviewed and approved

PR #201, opened 2024-09-06
Add google tag manager

AdrianMachado:
Adding google tag manager for conversion tracking


dan-lee (commented):
Reviewed and approved

PR #204, opened 2024-09-07
Publish prerelease versions without git commit

ntotten:
Changes the prerelease versions to use a version number `0.0.0-{github action run Id}` so that we don't need to commit the prerelease version each time.


lcampos (commented):
Reviewed and approved

PR #241, opened 2024-10-01
fix: Add back OpenAPI plugin path

dan-lee:
This was mistakenly removed but is needed.
Basic tests should cover this


lcampos (commented):
Reviewed and approved

PR #243, opened 2024-10-02
Prevent duplicate slugs

dan-lee:
Some APIs may have operations with same `summary` and `method`. This leads to duplicate IDs and duplicate active states in the sidebar.
To prevent this (and not have to add the `path` to the slug), I've added `slugifyWithCounter', which takes care of that.


packages/zudoku/src/vite/plugin-api.ts
```diff
@@ -28,18 +28,16 @@ const viteApiPlugin = (getConfig: () => ZudokuPluginOptions): Plugin => {
...
 
-              let data: unknown;
-              if (raw.trim().startsWith("{")) {
-                data = JSON.parse(raw);
-              } else {
+              if (!data.trim().startsWith("{")) {
                 const yaml = await import("yaml");
-                data = yaml.parse(raw);
+                data = yaml.parse(data);
               }
```
dan-lee, line 36: This was parsed and stringified again, which could result in a different sort order of `paths`

mosch (approved):

Issue #188, opened 2024-09-06
Support a Server switcher when OpenAPI spec includes multiple server 


zamd:
We have multiple regional servers which the user is able to switch but I don't see any option to do it with zudoku. Here is snippet of our OpenAPI yaml. 

```JSON
"servers": [
    {
      "description": "Beta Cluster",
      "url": "https://beta.authlete.com/"
    },
    {
      "description": "🇺🇸 US Cluster",
      "url": "https://us.authlete.com/"
    },
    {
      "description": "🇯🇵 Japan Cluster",
      "url": "https://jp.authlete.com/"
    },
    {
      "description": "🇪🇺 Europe Cluster",
      "url": "https://eu.authlete.com/"
    },
    {
      "description": "🇧🇷 Brazil Cluster",
      "url": "https://br.authlete.com/"
    }
  ]
  ```

PR #247, opened 2024-10-04
feat: Add OAS server selector in overview list and playground

dan-lee:
Addresses #188 

- [x] List servers in list overview
- [x] List servers in playground
- [x] Keep selection in sync

### DIFF
--- a/packages/zudoku/src/lib/plugins/openapi/playground/Playground.tsx
+++ b/packages/zudoku/src/lib/plugins/openapi/playground/Playground.tsx
@@ -256,10 +256,6 @@ export const Playground = ({
                 Send
               </Button>
             </div>
-            {/*            <label className="flex gap-2 items-center text-sm">
-              <input type="checkbox" />
-              Show host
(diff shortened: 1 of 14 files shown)


packages/zudoku/src/lib/plugins/openapi/playground/Playground.tsx
```diff
@@ -222,6 +256,10 @@ export const Playground = ({
                 Send
               </Button>
             </div>
+            {/*            <label className="flex gap-2 items-center text-sm">
```
mosch: Do we need to keep that around?

mosch (approved):
Great work, lets get it in.

packages/zudoku/src/lib/plugins/openapi/playground/Playground.tsx
dan-lee: Oops, no. That was just an idea, but don't think we need that right now

PR #258, opened 2024-10-08
Fix config reload

dan-lee:
Checks for import dependencies start with `file://` otherwise the protocol is duplicated (wouldn't happen on Windows systems)


lcampos (commented):
Reviewed and approved

PR #269, opened 2024-10-14
Package upgrades and cleanup

dan-lee:


scripts/package.json
```diff
@@ -6,11 +6,11 @@
...
   "dependencies": {
-    "@graphql-codegen/cli": "^5.0.2",
-    "prettier": "^3.3.3"
+    "@graphql-codegen/cli": "5.0.3",
+    "prettier": "3.3.3"
   },
   "devDependencies": {
-    "@types/node": "20",
-    "typescript": "5.5.3"
+    "@types/node": "22.7.5",
```
mosch: if we define 20 as requirement, maybe we should stick to the types

packages/zudoku/src/lib/components/index.ts
```diff
@@ -16,7 +16,7 @@ export const useMDXComponents = /*@__PURE__*/ useMDXComponentsImport;
 export const Callout = /*@__PURE__*/ CalloutImport;
 export const DevPortal = /*@__PURE__*/ DevPortalImport;
 export const Layout = /*@__PURE__*/ LayoutImport;
-export const Link = /*@__PURE__*/ LinkImport;
+export const Link: typeof LinkImport = /*@__PURE__*/ LinkImport;
```
dan-lee, line 19: > The inferred type of Link cannot be named without a reference to .../@types/react/

mosch (approved):

PR #272, opened 2024-10-16
Add component render to custom pages

dan-lee:
This adds the following to the `customPages` config/plugin

- `render`: allows to render components with exposed props passed in (same as in Slotlets)
- `prose`: So the custom page is wrapped in Tailwind prose class, not sure if this makes too much sense, but it was marked as `TODO` and thought I might just add it while working on it


mosch (approved):
At this point it feel it might be the right thing to expose a hook instead. But let's go for this now.

PR #279, opened 2024-10-17
Sidebar fixes

dan-lee:
- Fix top-level margins for docs
- Align sidebar items and badges properly
- Fix category open animation sometimes not firing correctly
- Allow docs to have no title: This was only so that the sidebar has a label to render (if nothing else is provided). IMO this is an unnecessary limitation and we can just fall back to the id (≈file name).


mosch (commented):
reviewed & approved

Issue #287, opened 2024-10-22
SSR: path.join breaks client entry path with custom base configuration


arian0zen:
## Issue: Incorrect Path Joining for Client-Side Hydration

**Problem:**
When using `path.join` to construct the client entry path in a Vite application with a custom base path, client-side hydration fails. This prevents client-side JavaScript from loading, leading to non-functional click events and other interactivity.

**Current Behavior:**
```javascript
const proxiedEntryClientPath = path.join(
  vite.config.base,
  "/__z/entry.client.tsx"
);
app.use(proxiedEntryClientPath, async (_req, res) => {
  // ... handler code
});
```
This code fails to serve the client entry file correctly due to the incorrect path construction using `path.join`.

**Expected Behavior:**
```javascript
app.use("/__z/entry.client.tsx", async (_req, res, next) => {
  // ... handler code
});
```
Using the direct path approach works correctly.

**Root Cause:**
- `path.join` uses system-specific path separators (e.g., backslashes on Windows).
- `path.join` may incorrectly combine base paths with the entry path.
- URL paths require forward slashes regardless of the operating system.

**Steps to Reproduce:**
1. Start the development server.
2. Open the application in a web browser.
3. Observe that client-side JavaScript is not loading and the application is non-functional.
4. And you can notice in the browser that the source code entry.client.tsx is empty

**Environment:**
- **OS:** Win 11
- **Node.js version:** 20
- **Framework version:** Zudoku@0.13.3 

**This mostly broke from @0.13.0 to @0.13.1**

**Explanation of Path Join Issue:**
When `path.join` is used with a base path of `/` and an entry path of `/__z/entry.client.tsx`, it incorrectly constructs the path as `\__z\entry.client.tsx`. This happens because `path.join` treats the base path as a relative path and attempts to join it with the entry path using the system's path separator (which is typically a backslash on Windows). However, URL paths require forward slashes, so this constructed path is invalid and prevents the client entry file from being served correctly.

**Solution:**
To avoid this issue, We can modify the path joining as:
```javascript
const proxiedEntryClientPath =
  (vite.config.base.endsWith("/")
    ? vite.config.base
    : vite.config.base + "/") + "__z/entry.client.tsx";
```
This ensures that the correct URL path is used to serve the client entry file, enabling proper client-side hydration.

dan-lee:
Should be fixed in 0.13.4. Thanks @arian0zen 👏

PR #288, opened 2024-10-22
fix: prevent path.join from creating malformed URL for client entry path

arian0zen:
When using path.join to construct the client entry path in a Vite application with a custom base path, client-side hydration fails. This prevents client-side JavaScript from loading, leading to non-functional click events and other interactivity.

This PR updates the joining method to fix this issue!

fiixes #287

### DIFF
--- a/packages/zudoku/src/vite/dev-server.ts
+++ b/packages/zudoku/src/vite/dev-server.ts
@@ -49,8 +49,14 @@ export class DevServer {
         ? vite.config.base
         : vite.config.base + "/") + "__z/entry.client.tsx";
 
+    // Ensure the path uses forward slashes, even on Windows
+    const normalizedProxiedEntryClientPath = proxiedEntryClientPath.replace(
+      /\\/g,
+      "/",
+    );
+
     app.use(graphql.graphqlEndpoint, graphql);
-    app.use(proxiedEntryClientPath, async (_req, res) => {
+    app.use(normalizedProxiedEntryClientPath, async (_req, res) => {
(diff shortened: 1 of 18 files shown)


mosch:
Awesome, thanks for the quick fix. And thank you for your contribution!

mosch (approved):

packages/zudoku/src/vite/dev-server.ts
```diff
@@ -44,12 +43,20 @@ export class DevServer {
...
+    const proxiedEntryClientPath =
+      (vite.config.base.endsWith("/")
+        ? vite.config.base
+        : vite.config.base + "/") + "__z/entry.client.tsx";
+
+    // Ensure the path uses forward slashes, even on Windows
+    const normalizedProxiedEntryClientPath = proxiedEntryClientPath.replace(
+      /\\/g,
+      "/",
     );
```
dan-lee: Wouldn't `path.posix.join` suffice?

arian0zen: ahh yeshh! That would be a better solution!
Updated the PR!

mosch:
We're struggleing with the pipeline, sorry. We merged it manually https://github.com/zuplo/zudoku/pull/302

arian0zen:
> We're struggleing with the pipeline, sorry. We merged it manually #302

is there any issue from my end ? shall I create a new branch in this repo itself rather than working on a new branch on the forked repo for the next feature that I might work on ?

PR #289, opened 2024-10-23
Adding option to keep the api categories closed by default -

### LINKED ISSUE #280: Sidebar Items by default closed
zudoku version 0.13.2

I want sidebar items should be closed by default, at present it keep opened.

is there any configuration or work around for this?
Screenshot from 2024-10-21 14-30-24

arian0zen:
- The section opens when a particular category is active.

[Here is how it looks - Loom video](https://www.loom.com/share/14ba9396fd764e0e81b2683cc5e7c834?sid=bfc6a143-85c1-486f-8c0a-24562c6f0c56)

- Added a key `defaultCollapsed` to the API config to determine if it should be collapsed by default.
- Try adding this key to the API config to test that.

image

### DIFF
--- a/packages/zudoku/src/config/validators/SidebarSchema.ts
+++ b/packages/zudoku/src/config/validators/SidebarSchema.ts
@@ -32,6 +32,7 @@ export type SidebarItemCategory = Omit<
   items: SidebarItem[];
   link?: SidebarItemCategoryLinkDoc;
   icon?: LucideIcon | string;
+  apiReference?: boolean;
 };
 
 export type SidebarItem =
--- a/packages/zudoku/src/lib/components/navigation/SidebarCategory.tsx
+++ b/packages/zudoku/src/lib/components/navigation/SidebarCategory.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import * as Collapsible from "@radix-ui/react-collapsible";
 import { ChevronRightIcon } from "lucide-react";
 import { useEffect, useMemo, useState } from "react";
--- a/packages/zudoku/src/lib/components/navigation/SidebarCategory.tsx
+++ b/packages/zudoku/src/lib/components/navigation/SidebarCategory.tsx
@@ -1,10 +1,12 @@
+/* eslint-disable no-console */
 import * as Collapsible from "@radix-ui/react-collapsible";
 import { ChevronRightIcon } from "lucide-react";
-import { useEffect, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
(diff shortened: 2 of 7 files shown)


packages/zudoku/src/lib/components/navigation/SidebarCategory.tsx
```diff
@@ -1,10 +1,12 @@
+/* eslint-disable no-console */
```
mosch: Is this a leftover we don't need anymore?

packages/zudoku/src/config/validators/SidebarSchema.ts
```diff
@@ -32,6 +32,7 @@ export type SidebarItemCategory = Omit<
   items: SidebarItem[];
   link?: SidebarItemCategoryLinkDoc;
   icon?: LucideIcon | string;
+  apiReference?: boolean;
```
mosch, line 35: Im not sure about the naming of this toggle. I think we should better describe the behaviour that we are toggleing here, like `openOnScroll`?

mosch:
Great feature! I'm thinking if this may be something that should be enabled / disabled on the first level of Zudoku.

packages/zudoku/src/lib/components/navigation/SidebarCategory.tsx
arian0zen: nyah, this is a miss out on my part, removing the line!

packages/zudoku/src/config/validators/SidebarSchema.ts
arian0zen: - the key apiReference is not actually user-facing, this one is just to identify if the collapsable categories are apiRef or not, if they are not apis and just documentation I figured we should not add the open on scroll ability.

- the open-on-scroll actually defined by the key defaultCollapsed, that one is user-facing. ( I think here this should be defined as `openOnScroll` )

- Although, I am very much open to this suggestion.. what do you say?

ntotten:

packages/zudoku/src/config/validators/SidebarSchema.ts
mosch: This is a Zudoku core type that should not know anything about the plugins using it. Also the flag controlls if the menu item automatically opens on scroll. So i think `openOnScroll` is the right way to call it.

mosch (changes requested):

mosch:
So @arian0zen with the current work on the schema we realized we want to build something similar soon. So let's close this for now. However our solution will be inspired by this - so thank you!

PR #293, opened 2024-10-27
Fix navigation show on mobile per display type

omurilo:
This PR fix the feature introduced on #290 on mobile devices, and add a feature to use `globalDisplay` for define the default behaviour of renderization to routes.

### DIFF
--- a/examples/with-auth0/zudoku.config.ts
+++ b/examples/with-auth0/zudoku.config.ts
@@ -1,6 +1,7 @@
 import { type ZudokuConfig } from "zudoku";
 
 const config: ZudokuConfig = {
+  globalDisplay: "auth",
   topNavigation: [
     { id: "documentation", label: "Documentation" },
     { id: "api", label: "Rick & Morty API" },
--- a/packages/zudoku/src/app/main.tsx
+++ b/packages/zudoku/src/app/main.tsx
@@ -31,6 +31,7 @@ export const convertZudokuConfigToOptions = (
     !config.page?.logo?.src?.dark;
 
   return {
+    globalDisplay: config.globalDisplay,
     page: {
       ...config.page,
       logo: {
--- a/packages/zudoku/src/config/validators/validate.ts
+++ b/packages/zudoku/src/config/validators/validate.ts
@@ -173,6 +173,7 @@ type BannerColorType = ZodOptional<
 const ConfigSchema = z
   .object({
     basePath: z.string().optional(),
+    globalDisplay: z.enum(["auth", "always"]).default("always").optional(),
     page: z
       .object({
         pageTitle: z.string(),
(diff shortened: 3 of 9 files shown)


mosch:
Hey @omurilo thanks for your contribution 👏 Great catch that the top-navigation item where not hidden on mobile yet. 

This PR contains a lot of changes, and some of them have very far-reaching effects. 
In order to get your changes in, lets move step by step. 

I created this PR https://github.com/zuplo/zudoku/pull/297 that only contains a fix to hide the items on mobile as well.

Other things is see in this PR:
- Introduce a `globalDisplay` to chagne the default behaviour
- Added missing openid auth type (should go in a seperate PR!)
- Hide all sidebar items if the top navigation item in hidden

Can you give us some details on the use-case so we can find an appropriate solution for want to achieve?

omurilo:
hey @mosch, ok no problem on separate the PR.

About openid types I can made a new PR for fix this, no problem.

About `globalDisplay` I think it a great introduction to make possible define the default behavior for more than one item and then set a anon `display` in an specific route for example.

About hide all sidebar items when top navigation is hidden, I think it a default behavior for this functionality, if I set a topNavigation display for the id `docs` for example, I don't want the user have access to the sidebar of `/docs`, don't make sense it view an item it he doesn't access.

mosch:
> About hide all sidebar items when top navigation is hidden, I think it a default behavior for this functionality, if I set a topNavigation display for the id docs for example, I don't want the user have access to the sidebar of /docs, don't make sense it view an item it he doesn't access.

I understand. This option is not ment do implement permission, but only to controll visiblity. It would be the responsibility of the page to implement the permssion. So this configuration should be on the page, not on the menu.

Can you tell me about what you're trying to build – so it would be easier for us to come up with a recommended solution!

omurilo:
> > About hide all sidebar items when top navigation is hidden, I think it a default behavior for this functionality, if I set a topNavigation display for the id docs for example, I don't want the user have access to the sidebar of /docs, don't make sense it view an item it he doesn't access.
> 
> 
> 
> I understand. This option is not ment do implement permission, but only to controll visiblity. It would be the responsibility of the page to implement the permssion. So this configuration should be on the page, not on the menu.
> 
> 
> 
> Can you tell me about what you're trying to build – so it would be easier for us to come up with a recommended solution!

If the permission is only on the page, the menu is always visible and this isn't the acceptable behavior in my opinion.

ntotten:

mosch:
Closing this for now – let us know if you want to get back to this.

mosch:
Update on the protectedRoutes: We drafted a feature that allows you to procted some routes with auth: https://github.com/zuplo/zudoku/pull/529

omurilo:
yeah guys, this cover my problem, nice draft.

mosch:
Thanks for the feedback @omurilo 👏

PR #305, opened 2024-10-28
Add authentication type `openid`

omurilo:
Today to use `openid` authentication I need to set `auth0` and add some different configs, so in this PR I added a validation config to authentication type `openid`.


dan-lee (approved):
Thank you 🙏

Issue #291, opened 2024-10-24
Add dropdowns to sandbox UI in place of text boxes when choices are constrained.


vengroff:
Many APIs have args that are in some sense strings but what they actually accept is essetially an enumerated type with only a fixed set of accepted values. In some cases they might accept a list of these items. In the sandbox, these appear as text fields. It would be more convenient for users, especially those trying out an API in the sandbox for the first time, if these were drop downs or multi-selects populated with the allowed values.

For reference, the sandbox-like feature that FastAPI generates in their docs pages does this. Examples look like the following:

![image](https://github.com/user-attachments/assets/50a3c7cd-8bbc-452e-bea1-2960a2e04641)
![image](https://github.com/user-attachments/assets/4bc5300e-001a-4204-a4be-17824004ef06)

In contrast, the zudoku sandbox for the same API that is shown in the first image above produces this:

<img width="663" alt="image" src="https://github.com/user-attachments/assets/58169e24-94b7-4612-a7b5-37380a50d153">

A drop down instead of a text box here would be great.

The code that generates the text input can be found at https://github.com/zuplo/zudoku/blob/main/packages/zudoku/src/lib/plugins/openapi/playground/QueryParams.tsx#L83. It would be great if this could be extended to use metadata from the OpenAPI spec to populate drop downs or multi-selects as appropriate.

vengroff:
Here are the relevant bits of the OpenAPI spec for the API shown above where the enum is defined and used:

```
      "RangeFuelType": {
        "type": "string",
        "enum": [
          "natural_gas",
          "propane"
        ],
        "title": "RangeFuelType",
        "description": "Fuel types inputs that the range upgrade API supports.\n\nNote that we do not support electricity as the input because we are modeling\nupgrading to electric, and so upgrading from electric would produce no\nchange."
      },
 ```
 
 and 
 
 ```
         "parameters": [
          {
            "name": "current_fuel",
            "in": "query",
            "required": true,
            "schema": {
              "$ref": "#/components/schemas/RangeFuelType",
              "title": "Current Range Fuel",
              "description": "The fuel used by the current range."
            },
            "description": "The fuel used by the current range."
          },
 ```

arian0zen:
@mosch @AdrianMachado 
am working on this one as well!

vengroff:
> @mosch @AdrianMachado am working on this one as well!

So this is work in progress? If so, that's awesome. Any ETA?

arian0zen:
[...]

PR #306, opened 2024-10-28
Replace Text Fields with Dropdowns for Enum Parameters in API Sandbox UI

arian0zen:
This PR introduces dropdowns and multi-selects in the Zudoku sandbox UI for parameters with constrained values, using the enum metadata from the OpenAPI spec. Instead of displaying text fields for these parameters, the updated UI now automatically populates dropdowns or multi-selects, streamlining the experience for users exploring the API.

This enhancement addresses issue [ZUP-3767](https://github.com/zuplo/zudoku/issues/291).

🎥 [Watch the Demo](https://www.loom.com/share/3f9f7a112cb14b8d8b4eead04a5b6d7f?sid=d70799ec-c2d6-4b21-8c3a-f1aa0360c640)

- added a custom MultiSelect.tsx component ( as Radix react select does not support multi-select primitive in existing dropdown )


vengroff:
Awesome. Thanks for making quick work of this issue.

dan-lee:
Thanks for your contribution @arian0zen! Looks great so far 🙏  Can you provide the changes you made to the OpenAPI schema so we can test it?

(and sorry for the merging all the time, the CI doesn't work properly with forks yet)

arian0zen:
> Thanks for your contribution @arian0zen! Looks great so far 🙏 Can you provide the changes you made to the OpenAPI schema so we can test it?
> 
> (and sorry for the merging all the time, the CI doesn't work properly with forks yet)


- for testing you can :
  a. set a few enum for type string in query param schema (line 131)
  b. also you can set type array and set enum for its items (line 156-159)
  
  
  
[openapi.json](https://github.com/user-attachments/files/17572932/openapi.json)

🫱🏽‍🫲🏾 I am very much open to any suggestions, let me know if any changes are required!

packages/zudoku/src/lib/plugins/openapi/playground/QueryParams.tsx
```diff
@@ -82,19 +90,67 @@ export const QueryParams = ({
...
+                              value={field.value as string}
+                              defaultValue={field.value as string}
+                            >
+                              <SelectTrigger
+                                className="w-full flex"
+                                placeholder="Choose an option"
+                                value={field.value}
+                              >
+                                <SelectValue />
+                              </SelectTrigger>
```
dan-lee, line 130: Let's just use `placeholder` that `SelectValue` provides. Then we don't need to add this prop to `SelectTrigger`
So:

```suggestion
                              <SelectTrigger
                                className="w-full flex"
                                value={field.value}
                              >
                                <SelectValue placeholder="Choose an option" />
                              </SelectTrigger>
```

packages/zudoku/src/lib/plugins/openapi/playground/Playground.tsx
```diff
@@ -198,13 +206,23 @@ export const Playground = ({
...
+        p.value.map((v, vi) => (
+          <Fragment key={`${p.name}-${v}`}>
+            {p.name}={encodeURIComponent(v)}
+            {(vi < p.value.length - 1 || i < arr.length - 1) && "&"}
+            <wbr />
+          </Fragment>
+        ))
+      ) : (
+        <Fragment key={p.name}>
+          {p.name}={encodeURIComponent(p.value)}
```
dan-lee, line 220: I think we should leave the replacement for `%20` to `+` in.

dan-lee (commented):
Awesome work! Thanks for the deep dive into the code base and the changes you’re proposing!
Some things:

- With the select fields now, when we select/change the values the table columns jump and change the layout. Can we avoid this?
- The multi select is a tricky component that is hard to get exactly right, e.g., some things I noticed:
    - When selecting a value, the dialog closes. I think it should stay open
    - I can’t unselect the last item if there’s only one item selected
    - A selected item should not have a background color because you can’t see hovered/active item. The checkmark should suffice
- Maybe we can get some inspiration from here: https://github.com/shadcn-ui/ui/issues/66#issuecomment-2254196754. E.g. this implementation: https://shadcnui-expansions.typeart.cc/docs/multiple-selector. 

I don’t want to block this but I think if we add such a versatile component it should be solid :)
Because we want to expose the components to be used by Zudoku users (in Markdown, custom pages, etc.)

arian0zen:
@dan-lee these are very solid suggestions. Working on it🙌

vengroff:
Hi,

Thanks for putting this code together.

I just wanted to check back in on the status and when it might be mergable. We could really use this for some APIs we are working on. 

Let me know if there is anything I can do to help.

Thanks!

ntotten:

mosch:
Done in https://github.com/zuplo/zudoku/pull/564

PR #313, opened 2024-11-04
Purge all CF cache when doing a release

dan-lee:
See https://developers.cloudflare.com/api-next/resources/cache/methods/purge/


ntotten (commented):
reviewed an approved

PR #317, opened 2024-11-05
expose auth for external to use

mosch:
This exposes the authentication state to be used.

### DIFF
--- a/packages/zudoku/src/lib/components/index.ts
+++ b/packages/zudoku/src/lib/components/index.ts
@@ -1,6 +1,7 @@
 import { useMDXComponents as useMDXComponentsImport } from "@mdx-js/react";
 import { Helmet } from "@zudoku/react-helmet-async";
 import { Link as LinkImport } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


packages/zudoku/src/lib/components/index.ts
```diff
@@ -12,16 +13,20 @@ import {
 import { DevPortal as DevPortalImport } from "./DevPortal.js";
 import { Layout as LayoutImport } from "./Layout.js";
 import { useZudoku as useZudokuImport } from "./context/ZudokuContext.js";
-
-export const useMDXComponents = /*@__PURE__*/ useMDXComponentsImport;
-export const Callout = /*@__PURE__*/ CalloutImport;
+export const useMDXComponents = /*@__P`URE__*/ useMDXComponentsImport;
```
dan-lee:
```suggestion
export const useMDXComponents = /*@__PURE__*/ useMDXComponentsImport;
```

dan-lee (approved):

PR #319, opened 2024-11-06
Use suspense + react query

mosch:
This PR changes to use react query over a custom implementation:
 - ✅ The callback gets executed only once, like before
 - ✅ It now uses the global suspense to show the loader
 - ✅ It now uses the global error page with the developer hint
 - ✅ The page is much easier to build
 - ✅ We don't have to use any on the error page


dan-lee (approved):
Great 👏

PR #321, opened 2024-11-08
Zudoku UI

dan-lee:
Adds commonly used shadcn components by default. Can be imported like:

```tsx
import { Button } from "zudoku/ui/Button";
import {
  Card,
  CardContent,
  CardDescription,
  CardHeader,
  CardTitle,
} from "zudoku/ui/Card";
```


mosch (approved):
👏 awesome

PR #332, opened 2024-11-15
Only create github release on tagged

ntotten:
Only create github releases on tagged versions.


lcampos (commented):
Reviewed and approved

PR #339, opened 2024-11-17
Set link-workspace-packages=true

ntotten:


lcampos (commented):
Reviewed and approved

PR #351, opened 2024-11-19
Fix `zudoku build` on Win

dan-lee:


lcampos (commented):
Reviewed and approved

PR #352, opened 2024-11-19
Fix GraphQL return type

dan-lee:


lcampos (commented):
Reviewed and approved

PR #361, opened 2024-11-20
Optimize off-screen content rendering using `content-visibility`

mosch:
This uses the cutting edge css feature to allow modern browser to render much faster.

### DIFF
--- a/packages/zudoku/src/lib/plugins/openapi/OperationList.tsx
+++ b/packages/zudoku/src/lib/plugins/openapi/OperationList.tsx
@@ -1,4 +1,5 @@
 import { ResultOf } from "@graphql-typed-document-node/core";
+import type { CSSProperties } from "react";
 import { useQuery } from "urql";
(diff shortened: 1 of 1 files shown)


packages/zudoku/src/lib/plugins/openapi/OperationList.tsx
```diff
@@ -154,7 +155,15 @@ export const OperationList = () => {
...
-          <div key={tag.name}>
+          <div
+            key={tag.name}
+            style={
+              {
+                "content-visibility": "auto",
+              } as unknown as CSSProperties
+              // we need this until we have types
+            }
+          >
```
dan-lee: We could also use Tailwind's [arbitrary properties](https://tailwindcss.com/docs/adding-custom-styles#arbitrary-properties).


```suggestion
          <div key={tag.name} className="[content-visibility:auto]">
```

dan-lee (approved):
Awesome 👏!

PR #371, opened 2024-11-25
Create new Github pre-release instead of updating it

dan-lee:
Release notes are not regenerated when updating the release so just delete the old one and create a new one.


ntotten (commented):
review and approved

PR #390, opened 2024-11-27
Prevent infinite loop

dan-lee:
This could result in creating infinite `QueryClients` but only in static generated files 🤯


ntotten (commented):
reviewed and approved

PR #403, opened 2024-12-06
Add `removeExtensions` and `removePaths` processors

dan-lee:
Building upon https://github.com/zuplo/zudoku/pull/402

Example 

```ts
import type { ZudokuConfig } from "zudoku";
import { removeExtensions } from "zudoku/post-processors/removeExtensions";
import { removePaths } from "zudoku/post-processors/removePaths";

const config: ZudokuConfig = {
  apis: {
    postProcessors: [
      removeExtensions({ keys: ["x-zuplo-route", "x-zuplo-path"] }),
      removePaths({
        paths: {
          // removes entire path:
          "/remove-me": true,

          // removes specific methods for path:
          "/remove-other": ["post"],
        },
        // custom filter (method is `true` for all methods)
        shouldRemove: ({ path, method, operation }) => operation["x-internal"],
      }),
    ],
    type: "file",
    navigationId: "api",
    input: "./openapi.json",
  },
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -40,7 +40,7 @@ jobs:
 
       - uses: nrwl/nx-set-shas@v4
 
-      - run: nx run-many -t root:lint:ci root:format:ci build:ci
+      - run: nx run-many -t root:lint:ci root:format:ci build:ci test
 
       - name: Get Variables
         id: vars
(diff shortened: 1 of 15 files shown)


ntotten (commented):
Reviewed and approved

PR #425, opened 2024-12-11
Support metadata on API reference page

AdrianMachado:
## Summary

Currently - the API reference pages do not have any metadata. This PR uses the OpenAPI spec's `info` property to add the title and description metatags. Perhaps we will want to allow overriding via `ZudokuConfig` but I feel like this is a pretty sensible default.


dan-lee (approved):
Awesome!

PR #427, opened 2024-12-12
Fix base path redirect

dan-lee:
Missing header `Location`


lcampos (commented):
Reviewed and approved

PR #430, opened 2024-12-13
Fix sidebar multi level indentation

dan-lee:


lcampos (commented):
Reviewed and approved

PR #447, opened 2024-12-19
fix docs upload path

ntotten:


dan-lee (commented):
Reviewed and approved

PR #448, opened 2024-12-19
Fixes lucide icon type

ntotten:
Currently the icon type does not work on external project. This generates a type that will work externally.

CleanShot 2024-12-19 at 07 52 35@2x


dan-lee (commented):
Reviewed and approved

PR #456, opened 2024-12-20
Support local .env files in config + react

ntotten:
Adds support for loading environment variables via `.env` files.


dan-lee (approved):
👏

PR #465, opened 2024-12-27
Fix footer `hyperlinks` to open in a new tab

shivam-sharma7:


shivam-sharma7:
@ntotten

ntotten:
We generally don't put _blank on URLs. While this is of course just personal preference, I prefer leaving them off so users can choose either control click or click.

PR #466, opened 2024-12-27
Include react-markdown in code rather than dep

ntotten:


vazexqi (commented):
Reviewed and approved

PR #469, opened 2024-12-27
Use Inkeep from CDN

ntotten:
The inkeep widget is huge and it has tons of dependencies that conflict with libraries we use. For these types of plugins, I think its better to load them from CDNs rather than bundle them.


vazexqi (commented):
Reviewed and approved

PR #505, opened 2025-01-09
Return empty array instead of undefined for standalone

dan-lee:
Fix ZUP-4017


mosch (commented):
Reviewed & Approved

ntotten (commented):
reviewed and approved

Issue #513, opened 2025-01-12
Zudoku build failing locally


shivam-sharma7:
I followed `Contributing.md`

```bash
PS C:\Users\shivams\Desktop\open-source\zudoku> nx run zudoku:build

> nx run zudoku:"generate:icon-types"

> zudoku@0.23.7 generate:icon-types C:\Users\shivams\Desktop\open-source\zudoku\packages\zudoku
> tsx ./scripts/generate-icon-types.ts
node:fs:442
    return binding.readFileUtf8(path, stringToFlags(options.flag));
                   ^
Error: ENOENT: no such file or directory, open 'C:\C:\Users\shivams\Desktop\open-source\zudoku\node_modules\.pnpm\lucide-react@0.468.0_react@19.0.0\node_modules\lucide-react\dynamicIconImports.js'
    at readFileSync (node:fs:442:20)
    at <anonymous> (C:\Users\shivams\Desktop\open-source\zudoku\packages\zudoku\scripts\generate-icon-types.ts:7:21)
    at ModuleJob.run (node:internal/modules/esm/module_job:271:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:547:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'C:\\C:\\Users\\shivams\\Desktop\\open-source\\zudoku\\node_modules\\.pnpm\\lucide-react@0.468.0_react@19.0.0\\node_modules\\lucide-react\\dynamicIconImports.js'
}
Node.js v22.13.0

```

shivam-sharma7:
@dan-lee @ntotten I have a fix that works, and resolve path currectly for Windows system.

PR #514, opened 2025-01-13
fix(generate-icon-types): correct path resolution for Windows base system

shivam-sharma7:
Correct path resolution for Windows base system for `dynamicIconImports.js`

Fixes: #513

## Before fix

Check issue

## After fix

```bash
PS C:\Users\shivams\Desktop\open-source\zudoku> nx run zudoku:build                                                    

> nx run zudoku:"generate:icon-types"

> zudoku@0.23.7 generate:icon-types C:\Users\shivams\Desktop\open-source\zudoku\packages\zudoku
> tsx ./scripts/generate-icon-types.ts

> nx run zudoku:build

> zudoku@0.23.7 build C:\Users\shivams\Desktop\open-source\zudoku\packages\zudoku
> tsc --project tsconfig.json

—————————————————————————————————————————————————————————————————————————————————————————————————————————————————————— 

 NX   Successfully ran target build for project zudoku and 1 task it depends on (25s)

View logs and investigate cache misses at https://nx.app/runs/5XCHpMaQ8t
```

### DIFF
--- a/packages/zudoku/scripts/generate-icon-types.ts
+++ b/packages/zudoku/scripts/generate-icon-types.ts
@@ -1,12 +1,10 @@
 import { readFileSync, writeFileSync } from "node:fs";
-import { createRequire } from "node:module";
-import path from "node:path";
 import { fileURLToPath } from "node:url";
 import { format } from "prettier";
 import * as ts from "typescript";
 
-const require = createRequire(import.meta.url);
-const filePath = require.resolve("lucide-react/dynamicIconImports.js");
+const path = import.meta.resolve("lucide-react/dynamicIconImports.js");
+const filePath = fileURLToPath(new URL(path));
 const fileContent = readFileSync(filePath, "utf-8");
 
 const sourceFile = ts.createSourceFile(
@@ -30,9 +28,7 @@ visit(sourceFile);
 let typeDefinition = `export type IconNames = ${iconNames.map((name) => `"${name}"`).join(" | ")};`;
 typeDefinition = await format(typeDefinition, { parser: "typescript" });
 
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const outputPath = path.resolve(
-  __dirname,
-  "../src/config/validators/icon-types.ts",
+const outputPath = fileURLToPath(
+  new URL("../src/config/validators/icon-types.ts", import.meta.url),
 );
 writeFileSync(outputPath, typeDefinition);
--- a/packages/zudoku/scripts/generate-icon-types.ts
+++ b/packages/zudoku/scripts/generate-icon-types.ts
@@ -1,9 +1,12 @@
 import { readFileSync, writeFileSync } from "node:fs";
+import { createRequire } from "node:module";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
 import { format } from "prettier";
 import * as ts from "typescript";
 
-const path = import.meta.resolve("lucide-react/dynamicIconImports.js");
-const filePath = new URL(path).pathname;
+const require = createRequire(import.meta.url);
+const filePath = require.resolve("lucide-react/dynamicIconImports.js");
 const fileContent = readFileSync(filePath, "utf-8");
 
 const sourceFile = ts.createSourceFile(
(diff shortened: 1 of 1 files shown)


packages/zudoku/scripts/generate-icon-types.ts
```diff
@@ -1,9 +1,12 @@
...
+import { createRequire } from "node:module";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
 import { format } from "prettier";
 import * as ts from "typescript";
 
-const path = import.meta.resolve("lucide-react/dynamicIconImports.js");
-const filePath = new URL(path).pathname;
+const require = createRequire(import.meta.url);
+const filePath = require.resolve("lucide-react/dynamicIconImports.js");
```
dan-lee: Why is this change necessary? What's the difference between `import.meta.resolve` and `require.resolve`?

shivarm: > Why is this change necessary? What's the difference between `import.meta.resolve` and `require.resolve`?

The change from using `import.meta.resolve` to `require.resolve` is necessary because `import.meta.resolve` is not a standard feature in Node.js, whereas `require.resolve` is a well-supported method for resolving module paths in CommonJS and ES module contexts.

dan-lee: That's not true: https://nodejs.org/docs/latest-v20.x/api/esm.html#importmetaresolvespecifier.

Can you check again if only the change at the `outputPath` variable might be enough?

shivarm: > That's not true: https://nodejs.org/docs/latest-v20.x/api/esm.html#importmetaresolvespecifier.

@dan-lee Please check this what i used: https://nodejs.org/docs/latest-v20.x/api/module.html

> Can you check again if only the change at the `outputPath` variable might be enough?

Okay

shivarm: >Can you check again if only the change at the `outputPath` variable might be enough?

Not enough. See the following err

```bash
PS C:\Users\shivams\Desktop\open-source\zudoku> nx run zudoku:build                                                    

> nx run zudoku:"generate:icon-types"

> zudoku@0.23.7 generate:icon-types C:\Users\shivams\Desktop\open-source\zudoku\packages\zudoku
> tsx ./scripts/generate-icon-types.ts
node:fs:442
    return binding.readFileUtf8(path, stringToFlags(options.flag));
                   ^
Error: ENOENT: no such file or directory, open 'C:\C:\Users\shivams\Desktop\open-source\zudoku\node_modules\.pnpm\lucide-react@0.468.0_react@19.0.0\node_modules\lucide-react\dynamicIconImports.js'
```

dan-lee:
```suggestion
const path = import.meta.resolve("lucide-react/dynamicIconImports.js");
const filePath = fileURLToPath(new URL(path));
```

packages/zudoku/scripts/generate-icon-types.ts
```diff
@@ -1,9 +1,12 @@
 import { readFileSync, writeFileSync } from "node:fs";
+import { createRequire } from "node:module";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
```
dan-lee:
```suggestion
import { fileURLToPath } from "node:url";
```

packages/zudoku/scripts/generate-icon-types.ts
```diff
@@ -27,8 +30,9 @@ visit(sourceFile);
...
 typeDefinition = await format(typeDefinition, { parser: "typescript" });
 
-const outputPath = new URL(
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const outputPath = path.resolve(
+  __dirname,
   "../src/config/validators/icon-types.ts",
-  import.meta.url,
-).pathname;
+);
```
dan-lee: This can just be:

```ts
const outputPath = fileURLToPath(
  new URL("../src/config/validators/icon-types.ts", import.meta.url),
);
```

(I can't put review suggestions on deleted lines)

dan-lee (changes requested):
I'm still not sure why `createRequire` is needed for that 🤔  Check my suggestions, I think just by using `fileURLToPath` it should be fixed

shivam-sharma7:
> I'm still not sure why `createRequire` is needed for that 🤔 Check my suggestions, 

@dan-lee No problem let's remove it. 

>I think just by using `fileURLToPath` it should be fixed

Yeah, It working, let's go with your solution

dan-lee (approved):

PR #515, opened 2025-01-13
feat: automatically fire browser on server start

shivam-sharma7:
Zudoku should able to fire browser automatically


shivam-sharma7:
@dan-lee feel free edit and give feedback

dan-lee (commented):
We use `port` in the Express listener, so I don't think we should pass it down to Vite config:

https://github.com/zuplo/zudoku/blob/6614f678053b925bc95ed6b60b5e414a1ac883fa/packages/zudoku/src/vite/dev-server.ts#L133-L134

PR #529, opened 2025-01-15
Allow to protect routes with login

mosch:


packages/zudoku/src/lib/core/ProtectedRoute.tsx
```diff
@@ -0,0 +1,35 @@
+import { useEffect } from "react";
+import { matchPath, useLocation } from "react-router";
+import { useAuth } from "../authentication/hook.js";
+import { useZudoku } from "../components/context/ZudokuContext.js";
+import { ZudokuError } from "../util/invariant.js";
+
+export function ProtectedRoute({ children }: { children: React.ReactNode }) {
```
dan-lee: Can we name it `RouteGuard` or something? It's a bit confusing in `main.tsx` that it's wrapped around the whole app.

packages/zudoku/src/lib/core/ProtectedRoute.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  if (isProtected && !auth.isAuthEnabled) {
+    throw new ZudokuError("Authentication is not enabled", {
+      title: "Authentication is not enabled",
+      developerHint:
+        "To use protectedRoutes you need authentication to be enabled",
+    });
+  }
+
+  return children;
```
dan-lee: Let's return an `<Outlet />` so we can just use this component in `main.tsx`

dan-lee (approved):
Good! some small remarks

PR #558, opened 2025-01-20
Fixes incorrect oidc token exchange redirect url

ntotten:
The redirec_uri must be the same as the callback url on token exchange


dan-lee (commented):
Reviewed and approved

PR #534, opened 2025-01-16
fix(vite): ensures the paths are correctly formatted for all os

shivam-sharma7:
paths should be correctly formatted, and the import should resolve without errors on all operating systems. This ensures compatibility and avoids issues related to incorrect path formatting

Fixes: #511

PR #562, opened 2025-01-21
Fix `schemaPath` in Windows

dan-lee:
cc @shivarm


shivarm:
@dan-lee then what about #534 ?

dan-lee:
Sorry, I didn't see you already rebased it

PR #568, opened 2025-01-21
Add canonical URL

AdrianMachado:
## Summary

Docs pages should have canonical URLs so that google does not accidentally index a duplicate version of the same page (ex. the same page with query params). This currently affects our docs.

### Question

Is there any way to know what domain we are on in `MdxPage.tsx` without using `window`? It's likely better to have the canonical URL in the `head` during SSR rather than waiting for CSR


AdrianMachado:
> Great! Today we don't have a way of telling what domain Zudoku will run on. However I see how this might be helpful in other places, so we should consider adding an optional configuration for the domain to the config.

Yeah I agree that would be useful. I'm going to keep the scope of this change small and merge it in to unbreak existing portals

dan-lee (commented):
Reviewed and approved

PR #573, opened 2025-01-22
style: Update text size from 'text-xs' to 'text-sm' in SidecarExample…

mosch:
…s component


dan-lee:
If we do this we should be consistent with the size changes in the sidecar, it looks a bit over the place now.

For me this change makes it look too bulky.

PR #575, opened 2025-01-22
Revert sidebar category changes

dan-lee:
We need indentation and I think we also should have the lines as guides which level.
See this example which actually has two levels:

CleanShot 2025-01-22 at 11 00 41


mosch:
Maybe the prev change was taking it too far, however I don't think the line works with the new line-layout. It's too many lines.
CleanShot 2025-01-22 at 15 12 57

How about some thing in between like a bit of indentation without a line?
CleanShot 2025-01-22 at 15 14 30

mosch (approved):

PR #577, opened 2025-01-22
Allow plugins to purge cache

mosch:


packages/zudoku/src/lib/oas/graphql/index.ts
```diff
@@ -463,7 +464,10 @@ builder.queryType({
           const module = await loadSchema();
           schema = module.schema;
         } else {
-          schema = await validate(args.input as string);
+          schema = await validate(
+            args.input as string,
+            args.baseDir ?? undefined,
+          );
```
dan-lee: Is this by mistake?

packages/zudoku/src/lib/components/index.ts
```diff
@@ -15,7 +15,9 @@ import { Layout as LayoutImport } from "./Layout.js";
 import { Markdown as MarkdownImport } from "./Markdown.js";
 import { Spinner as SpinnerImport } from "./Spinner.js";
 import { Zudoku as ZudokuImport } from "./Zudoku.js";
+import { useCache as useCacheImport } from "./cache.js";
```
dan-lee: We need to export `CACHE_KEYS` as well

dan-lee (changes requested):

dan-lee (approved):

Issue #579, opened 2025-01-23
Migration `tailwindcss` v3 to v4 to optimize build performence


shivarm:
Here what i like about this:

- [New high-performance engine](https://tailwindcss.com/blog/tailwindcss-v4#new-high-performance-engine)
- [Simplified installation](https://tailwindcss.com/blog/tailwindcss-v4#simplified-installation)
- [CSS theme variables](https://tailwindcss.com/blog/tailwindcss-v4#css-theme-variables)

Take a look: https://tailwindcss.com/blog/tailwindcss-v4#new-high-performance-engine

@dan-lee @mosch A draft PR is opened

PR #580, opened 2025-01-23
website: migrate tailwindcss v3 to v4

shivarm:
For more information check issue

closes: #579


shivarm:
@dan-lee I need your help here https://github.com/zuplo/zudoku/blob/34f0b3d65e4dfcb3b781597c4d069a6edbd550bc/packages/zudoku/package.json#L239

In order to migrate, we have to go through the config and you know better than me. So it would be nice if you can take a look

dan-lee:
> @dan-lee I need your help here
> 
> https://github.com/zuplo/zudoku/blob/34f0b3d65e4dfcb3b781597c4d069a6edbd550bc/packages/zudoku/package.json#L239
> 
> In order to migrate, we have to go through the config and you know better than me. So it would be nice if you can take a look

Hey @shivarm I think that transition isn't as easy, as we configure a lot of Tailwind in the PostCSS plugin. Going configless will create some challenges. We probably need to use [`@source`](https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-registering-sources) but this will be a bit tricky as the sources are dynamic in our case. Maybe a Vite plugin that creates the sources dynamically in the `load` hook will do.

So basically code generation (as we do in a lot of Vite plugins).

1. Import like `virtual:zudoku-tailwind.css`
2. Register Vite plugin that hooks into that
3. Generate CSS text content with a bunch of `@source` calls that are put together similar to this:
https://github.com/zuplo/zudoku/blob/9713d55c0979da8f6d4a26392cb59eeeb58c147c/packages/zudoku/src/vite/config.ts#L303-L317

shivarm:
> > @dan-lee I need your help here
> > https://github.com/zuplo/zudoku/blob/34f0b3d65e4dfcb3b781597c4d069a6edbd550bc/packages/zudoku/package.json#L239
> > 
> > In order to migrate, we have to go through the config and you know better than me. So it would be nice if you can take a look
> 
> Hey @shivarm I think that transition isn't as easy, as we configure a lot of Tailwind in the PostCSS plugin. Going configless will create some challenges. We probably need to use [`@source`](https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-registering-sources) but this will be a bit tricky as the sources are dynamic in our case. Maybe a Vite plugin that creates the sources dynamically in the `load` hook will do.
> 
> So basically code generation (as we do in a lot of Vite plugins).
> 
> 1. Import like `virtual:zudoku-tailwind.css`
> 2. Register Vite plugin that hooks into that
> 3. Generate CSS text content with a bunch of `@source` calls that are put together similar to this:
>    https://github.com/zuplo/zudoku/blob/9713d55c0979da8f6d4a26392cb59eeeb58c147c/packages/zudoku/src/vite/config.ts#L303-L317

@dan-lee  If this is, then can we merge it for website?

dan-lee (approved):

dan-lee:
Thanks so much @shivarm! 🙏

PR #587, opened 2025-01-23
Add option `enableStatusPages` to generate http static pages

mosch:


packages/zudoku/src/lib/components/ErrorPage.tsx
```diff
@@ -13,7 +13,7 @@ export const ErrorPage = ({
   category?: ReactNode;
 }) => {
   return (
-    <div className={ProseClasses + " h-full pt-[--padding-content-top]"}>
+    <div className={ProseClasses + "h-full pt-[--padding-content-top]"}>
```
dan-lee: Shouldn't remove the space, or use `cn`

dan-lee (changes requested):

dan-lee (approved):

PR #606, opened 2025-01-27
Fix running format:ci, lint:ci in action

ntotten:
Formatting and linting check wasn't running correctly in the actions.


packages/zudoku/src/app/demo-cdn.html
dan-lee, line 1: I don't think this file should be formatted

PR #623, opened 2025-01-30
Fix circular structure detection

dan-lee:


mosch:
CleanShot 2025-01-30 at 11 25 34@2x

I think this badge should use the same text size as the other.
eAlso it's a bit unclear what we are referring to. Any chance we can link to the anchor of the element or display a label?

dan-lee:
> I think this badge should use the same text size as the other.
One of your PR removes the badges again, so I agree that it looks a bit odd, but I think we should see how it looks when the badges are removed again? I'd like to get this fix in.

> eAlso it's a bit unclear what we are referring to. Any chance we can link to the anchor of the element or display a label?
It's self-referencing. I can try but it's a bit tricky to work with this. We don't have the schema name to work with currently, so we are limited in what we can actually do.

The most important part is that it fixes this issue, which I forgot to mention in the PRs description:

CleanShot 2025-01-30 at 11 43 08

mosch (approved):

PR #625, opened 2025-01-30
Fix wrong context resolution

dan-lee:
When using any React context in the `app` folder (`RouteGuard` in this case), we have to make sure that the same context is used in the distributed files, so our current solution/workaround is to re-export from `zudoku/components` and use it in `main.tsx`.

At some point we need to address this properly so that we don't run into it again by accident.


mosch (commented):
Reviewed & Approved

PR #627, opened 2025-01-31
Playground improvements

dan-lee:
- Header enums support
- Suggest header name based what's passed by the schema (if not already added)
- More alignment between all parameter types
- Focus on input field on path param click

You can check it in the playground of the [first endpoint of Shipments API](https://zudoku-cosmo-cargo-git-feat-playground-header-improvement.zuplosite.com/api-shipments/shipment-management#create-a-new-shipment-create-shipment)

### DIFF
--- a/packages/zudoku/src/lib/components/Autocomplete.tsx
+++ b/packages/zudoku/src/lib/components/Autocomplete.tsx
@@ -1,6 +1,6 @@
 import { PopoverAnchor } from "@radix-ui/react-popover";
 import { useCommandState } from "cmdk";
-import { useRef, useState, type Ref } from "react";
+import { useRef, useState, type KeyboardEvent, type Ref } from "react";
 import {
   Command,
   CommandInlineInput,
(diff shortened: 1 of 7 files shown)


packages/zudoku/src/lib/components/Autocomplete.tsx
```diff
@@ -104,7 +106,7 @@ const AutocompletePopover = ({
 
 export const Autocomplete = (props: AutocompleteProps) => {
   return (
-    <Command className="bg-transparent">
+    <Command className="bg-transparent" shouldFilter={props.shouldFilter}>
```
shivarm:
```suggestion
    <Command className="bg-transparent" shouldFilter={shouldFilter}>
```
@dan-lee use destructure instead `props.` first pass it in `AutocompletePopover` function

shivarm (changes requested):

packages/zudoku/src/lib/components/Autocomplete.tsx
dan-lee: Thanks! applied

PR #645, opened 2025-02-05
Ignore major upgrades by dependabot

shivarm:
many times we won't be able to upgrade to a major version due to deprecation of lib, or functions etc..

Also, take example of #635 this not an automated upgrade.


dan-lee:
I don't think we should ignore all major versions by default, but rather on a case-by-case basis (e.g. this is what we did for ESLint).
This gives us a heads-up that there's a new version available, and we can use this PR/branch to apply fixes for breaking changes.

PR #652, opened 2025-02-06
Better logging for prerendering

dan-lee:
- Better TTY logging
- Better non-TTY logging
- Get rid of some warnings
- Timeout when pre-rendering worker hangs for some reason


packages/zudoku/src/lib/plugins/markdown/resolver.ts
```diff
@@ -1,13 +1,8 @@
-import fs from "fs";
```
dan-lee, line 1: This gave us a warning in every build that these APIs are not available on client side. The method that uses them is not in use (anymore?)

ntotten: great find!!!!

ntotten (approved):

PR #656, opened 2025-02-08
fix: avoid blocking main thread with using `node:fs/promises`

shivarm:


packages/zudoku/src/vite/plugin-metadata.ts
```diff
@@ -18,11 +19,14 @@ const viteBuildMetadata = (): Plugin => {
...
-        path.join(__dirname, "graph.json"),
-        JSON.stringify(deps, null, 2),
-      );
+      try {
+        await writeFile(
+          path.join(__dirname, "graph.json"),
+          JSON.stringify(deps, null, 2),
+        );
+      } catch (error) {
+        logger.error("Error writing graph.json:", error);
```
mosch, line 28: We'd need to re-throw the error, otherwise the build will not fail, but it should.

mosch:
Hey @shivarm, thanks for this contribution. I tested it and I don't think it bringings a meaningful change to the performance of the scripts. So I'll close this for now.

shivarm:
@mosch 
- synchronous blocks the execution of other code until the file system operation completes
- promises promise that resolves or rejects when the file system operation completes, allowing other code to execute in the meantime

Especially those requiring scalability and responsiveness, the fs/promises API is recommended. It allows for efficient handling of asynchronous operations without blocking the main thread.

PR #670, opened 2025-02-12
Fix CDN url path join

dan-lee:
`new URL()` strips away the pathname of the [`base` parameter](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#base) but we explicitely allow it.


lcampos (commented):
Reviewed and approved

PR #701, opened 2025-02-18
Allow page to take full width for pages without TOC

mosch:


dan-lee:
I don't think we can just remove the `max-width`, e.g. [`Configuration`](https://zudoku-docs-git-toc-page-width.zuplopreview.net/docs/configuration/overview) is now slightly broken (maybr other pages too)

Can't we just increase the prose `max-width`? The default is `65ch` which might be too narrow, but we could up it to `80ch` or `100ch`.

CleanShot 2025-02-18 at 11 39 25

mosch:
@dan-lee yeah, mb. The idea was to give the content the room it has if there is no TOC. I fixed that so we can use the full space without TOC.

PR #704, opened 2025-02-18
build samples

mosch:


dan-lee:
We do this already here: https://github.com/zuplo/zudoku/blob/8a28fbe0f8b23e6c0e69ba7895f27cd1c18f7605/.github/workflows/build.yaml#L58-L59

PR #708, opened 2025-02-19
Allow to consume events from Zudoku

mosch:


docs/pages/custom-plugins.md
```diff
@@ -3,6 +3,113 @@ title: Custom Plugins
...
+### Event Consumer Plugin
+
+```tsx
+import { ZudokuPlugin } from "zudoku";
+
+const eventConsumerPlugin: ZudokuPlugin = {
+  events: {
+    onRouteChange: (path) => {
+      console.log(`Route changed to: ${path}`);
+    },
```
leggetter: Would be great if this had `from` and `to` because some analytics tools have a `$pageview` and `$pageleave` event e.g., PostHog. However, I'm not sure how often the leave event is actually used.

That said, maybe there are other use cases for this.

docs/pages/extending/events.md
```diff
@@ -0,0 +1,93 @@
...
+## Using Events in Plugins
+
+To consume events in your plugin, you can implement the events property in your plugin:
+
+```typescript
+import { ZudokuPlugin, ZudokuEvents } from "zudoku";
+
+const navigationLoggerPlugin: ZudokuPlugin = {
+  events: {
+    location: (location) => {
```
leggetter: Same as above: from and to would be good.

leggetter (commented):
In general, looks great. One suggestion in the comments.

docs/pages/custom-plugins.md
dan-lee: Good suggestion! I've change it to emit 

```ts
{ from?: Location, to: Location }
```

dan-lee: Oh this is a different document, I think these were examples from the beginning.

This is the actual signature: https://zudoku-docs-git-zudoku-events.zuplopreview.net/docs/extending/events#location

dan-lee (approved):

PR #709, opened 2025-02-19
Add documentation for theming

mosch:


docs/pages/components/syntax-highlight.mdx
dan-lee, line 1: I don't think this file is put in the sidebar navigation, we also should reference ['Code Blocks' documentiation](https://zudoku-docs-git-theme.zuplopreview.net/docs/markdown/code-blocks) and vice versa.

docs/pages/customization/available-variables.mdx
dan-lee, line 1: Weird file/path name for "Custom Theme" doc

dan-lee (commented):
Sometimes this happens

https://github.com/user-attachments/assets/10c14bca-23b8-4b3f-8d85-f38c4e4f3a63

dan-lee:
Radius 0 doesn't seem to work

https://github.com/user-attachments/assets/9743af80-3379-4761-beed-4809dfe28ed3

dan-lee:
Awesome! 👏 
I think the layout is a bit lacking though in comparison to shadcn for example. What if we had the copy code as a button with a modal similar as they do? I don't think we need to show the resulting config all the time

dan-lee (approved):

dan-lee (approved):

Issue #455, opened 2024-12-20
Upgrade to new `node.js 22.x` LTS


shivam-sharma7:
Currently project using node 20.x and node 22.x is new LTS and node.js team also saying that you should upgrade node.js and in new LTS enable many feature and supports.

upgrade area- https://github.com/zuplo/zudoku/blob/main/.tool-versions

also- https://github.com/zuplo/zudoku/blob/8ae70dc347d4d3ea28ec63e674e20fe1ba15ad30/package.json#L29

cc: @ntotten @dan-lee

I would love to open a PR

ntotten:
The project supports Node 20+. It should work already on Node 22. Have you encountered any issues?

shivam-sharma7:
@ntotten Node 22 has performance improvements and security enhancements

ntotten:
We aren't going to make Node 22 a requirement, but the project does run on Node 22 so can use node 22 on your project.

PR #738, opened 2025-02-24
Switch to Node.js 22

dan-lee:


mosch (approved):
General fine, maybe @ntotten has more thoughts?

ntotten:
yup, we already switched the zuplo build to use 22.

shivarm:
I already discussed here- #455

PR #744, opened 2025-02-25
Playground: Ask user to sign in if required

mosch:


packages/zudoku/src/lib/authentication/hook.ts
```diff
@@ -28,5 +32,12 @@ export const useAuth = () => {
       // Redirect to home
       window.location.href = "/";
     },
+
+    signup: async () => {
+      if (!isAuthEnabled) {
+        throw new Error("Authentication is not enabled.");
+      }
+      await authentication.signUp();
```
dan-lee:
```suggestion
      await authentication.signUp({
        redirectTo: window.location.href,
      });
```

dan-lee (approved):

PR #766, opened 2025-03-03
Playground: Ask for key when submitting

mosch:


packages/zudoku/src/lib/plugins/openapi/playground/IdentityDialog.tsx
```diff
@@ -0,0 +1,74 @@
...
+              }
+            />
+            <Label htmlFor="remember">Remember my choice</Label>
+          </div>
+
+          <Button
+            onClick={() => onSubmit({ identity: identity, rememberedIdentity })}
+          >
+            Send
+          </Button>
```
dan-lee, line 67: Should we force a selection before being able to send? Right now I am able to just hit send and even check "Remember my choice" without selecting an option, feels a bit weird

packages/zudoku/src/lib/plugins/openapi/playground/Playground.tsx
```diff
@@ -158,27 +164,28 @@ export const Playground = ({
...
-
-  const formRef = useRef<HTMLFormElement>(null);
+  }, [latestSetRememberedIdentity, formState.identity]);
 
   const queryMutation = useMutation({
     mutationFn: async (data: PlaygroundForm) => {
       const start = performance.now();
+
+      if (identities.isLoading) {
+        await identities.promise;
```
dan-lee: Ha, didn't know `.promise` before, sounds cool but also:
> Requires the experimental_prefetchInRender feature flag to be enabled.

<sub>https://tanstack.com/query/latest/docs/framework/react/guides/suspense#using-usequerypromise-and-reactuse-experimental</sub>

IIRC it's intended to be used with `React.use`. Not sure if it works by chance?

dan-lee (commented):
In general looks good, a few remarks

packages/zudoku/src/lib/plugins/openapi/playground/IdentityDialog.tsx
mosch: We're now defaulting to "no auth"

dan-lee (approved):

PR #784, opened 2025-03-06
Improve prerender performance

dan-lee:


packages/zudoku/src/app/entry.server.tsx
```diff
@@ -8,31 +8,31 @@ import {
...
 }: {
   template: string;
   request: express.Request | Request;
   response: express.Response | FileWritingResponse;
-  config: ZudokuConfig;
+  routes: RouteObject[];
+  basePath?: string;
 }) => {
-  const routes = getRoutesByConfig(config);
   const { query, dataRoutes } = createStaticHandler(routes, {
```
dan-lee, line 34: Passing`routes` in from the outside instead of running `getRoutesByConfig` is by far the most crucial performance gain

mosch (approved):

PR #788, opened 2025-03-06
Fix broken admonitions link

ntotten:


dan-lee (commented):
Reviewed and approved

PR #834, opened 2025-03-25
Fix create-app template

dan-lee:


packages/zudoku/src/vite/config.ts
```diff
@@ -181,10 +181,7 @@ export async function getViteConfig(
     envPrefix,
     resolve: {
       alias: {
-        "@mdx-js/react": path.resolve(
-          config.__meta.moduleDir,
-          "node_modules/@mdx-js/react",
-        ),
+        "@mdx-js/react": import.meta.resolve("@mdx-js/react"),
```
dan-lee, line 184: Example failed with `pnpm`

mosch (approved):

PR #841, opened 2025-03-27
Remove client side canonical URL

dan-lee:
This can be done by adding a plugin in the Zudoku config:

```ts
plugins: [{
  getHead: ({ location }) =>
    <link
      rel="canonical"
      href={"https://example.com/docs" + location.pathname}
    />
}]
```


dan-lee:
We've discussed this and I'm going to create a separate top-level config for this

docs/pages/configuration/overview.md
```diff
@@ -262,6 +262,16 @@ The port on which the development server will run. Defaults to `3000`. This opti
 
 If the port is already in use, the next available port will be used.
 
+### `canonicalUrl`
+
+Sets the canonical URL for your documentation site. This is used for SEO purposes and helps search engines understand the preferred version of a page.
+
+```json
+{
+  "canonicalUrl": "https://docs.example.com"
```
AdrianMachado: It would be good to give an example that shows that the true value is `canonicalUrl + basePath + location`

docs/pages/configuration/overview.md
```diff
@@ -262,6 +262,16 @@ The port on which the development server will run. Defaults to `3000`. This opti
 
 If the port is already in use, the next available port will be used.
 
+### `canonicalUrl`
```
AdrianMachado: Maybe call this `canonicalUrlOrigin` so its clear that this is just the [URL origin](https://developer.mozilla.org/en-US/docs/Web/API/URL/origin)

AdrianMachado (approved):

AdrianMachado (commented):
Beautiful!

PR #865, opened 2025-04-04
Add scheme titles to logical group display

LLMK:
Adds scheme titles to anyOf, oneOf and other groups where more than one scheme may exist.

I noticed these don't exist which makes it difficult to use endpoints with multiple bodies. See comparison below.

**Before**
image

**After**
image

For reference, here's how the swagger editor displays multiple schemas in a oneOf:
image


mosch:
Great change, Thank you! 👏

dan-lee (approved):
👏

PR #896, opened 2025-04-14
Added dynamic padding to sidebar to prevent the last elements to drown in the gradient

nocturnalpollinator:
## The problem
The gradient mask on sidebar has a dynamic height at the bottom:
```
maskImage: `linear-gradient(180deg, transparent 1%, rgba(0, 0, 0, 1) 20px, rgba(0, 0, 0, 1) 90%, transparent 99%)`,
```
The bottom padding on the nav element is fixed:
```
-mx-[--padding-nav-item] **pb-3** pt-[--padding-content-top] scroll-pt-2 gap-1
```

This causes the last element in a long sidebar to "drown" in the gradient if the window size is tall:
Screenshot 2025-04-14 at 10 13 26

## The solution
Added a dynamic height to the bottom padding using vh since padding percentages are based on the element width and not height.


dan-lee:
Awesome, thanks for your contribution @nocturnalpollinator!

dan-lee (approved):

PR #908, opened 2025-04-14
Trigger internal zudoku workflow on release

ntotten:
Triggers the internal zudoku builder workflow when we run a release.


lcampos (commented):
Reviewed and approved

PR #909, opened 2025-04-14
Fix Zuplo base processors

dan-lee:


lcampos (commented):
Reviewed and approved

PR #919, opened 2025-04-16
Remove mention of create-zudoku-app from docs

ntotten:
Removing the create-zudoku-app reference to keep the doc clean as this is used on the zuplo docs as well.


lcampos (commented):
Reviewed and approved

PR #924, opened 2025-04-17
Refactor schema management

dan-lee:
I tried to add a simple cache buster query param which led to this refactoring. The schema processing was a bit messy and I tried to make everything a bit more manageable

### DIFF
--- a/packages/zudoku/src/vite/plugin-api.ts
+++ b/packages/zudoku/src/vite/plugin-api.ts
@@ -1,5 +1,6 @@
 import fs from "node:fs/promises";
 import path from "node:path";
+import { deepEquals } from "nx/src/utils/json-diff.js";
 import { type Plugin, runnerImport } from "vite";
 import { type LoadedConfig } from "../config/config.js";
 import { fileExists } from "../config/loader.js";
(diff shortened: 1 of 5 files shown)


packages/zudoku/src/vite/plugin-api.ts
```diff
@@ -304,9 +191,9 @@ const viteApiPlugin = async (
               `    ...(apis[${apiIndex}].options ?? {}),`,
               `  },`,
               `  schemaImports: {`,
-              ...Array.from(schemaMap.entries()).map(
-                ([key, schemaPath]) =>
-                  `    "${key}": () => import("${schemaPath.replace(/\\/g, "/")}"),`,
+              ...schemaMapEntries.map(
+                ([key, processed]) =>
+                  `    "${key}": () => import("${processed.filePath.replace(/\\/g, "/")}?d=${processed.processedTime}"),`,
```
dan-lee, line 198: This adds a cache buster

packages/zudoku/src/vite/plugin-api.ts
```diff
@@ -166,34 +56,37 @@ const viteApiPlugin = async (
...
+        if (!schemaManager.trackedFiles.has(id)) return;
 
-        processedSchemas = await processSchemas(tmpDir, getConfig(), [
-          ...buildProcessors,
-          ...zuploProcessors,
-        ]);
-        allSchemaFiles.forEach((file) => server.watcher.add(file));
+        // eslint-disable-next-line no-console
+        console.log(`Re-processing schema ${id}`);
+        await schemaManager.processSchema(id);
```
dan-lee: Now single schemas can be reprocessed instead of every single one all the time

Issue #926, opened 2025-04-18
OAuth0 authentication doesn't work if base path is used


ashipilov:
Steps to reproduce:

1. Specify basePath `/docs` at `examples/with-auth0/zudoku.config.ts`
2. Run nx run with-auth0:dev;
3. Go to http://localhost:3000/docs/api/characters#get-all-characters and try to login

Expected result:
- You are logged in and redirected to http://localhost:3000/docs/api/characters#get-all-characters

Observed result:
- Auth0 redirects you back to the proper url but the authorization code is not picked up

ashipilov:
https://github.com/zuplo/zudoku/pull/927

Joren-Thijs-KasparSolutions:
Hi i just ran into the exact same problem. I see a fix is already merged. when will the release be available?

dan-lee:
@Joren-Thijs-KasparSolutions There's a new version [`v0.41.0`](https://github.com/zuplo/zudoku/releases/tag/v0.41.0) out now

PR #927, opened 2025-04-18
Fix for OpenID auth if base bath is used

ashipilov:
Fixes https://github.com/zuplo/zudoku/issues/926

Test Directions

1. Specify basePath `/docs` at `examples/with-auth0/zudoku.config.ts`
2. Run nx run with-auth0:dev;
3. Go to http://localhost:3000/docs/api/characters#get-all-characters and try to login

Successful result:
- You are logged in and redirected to http://localhost:3000/docs/api/characters#get-all-characters

Failed result:
- You are not logged in or not end up at the correct page


dan-lee (approved):
Thanks!

Issue #894, opened 2025-04-13
Improve error message for invalid config file


ntotten:
A configuration error resulting from a bad import logs "no zudoku config file found in project root.". The error should say something like "invalid configuration" and not to look at the earlier logs to see the error.

![Image](https://github.com/user-attachments/assets/5f7c7d21-1268-46f9-958d-73c9d2f640b5)

PR #953, opened 2025-04-30
Improve error message for invalid config file

shivarm:
fixes: #894


mosch:
Thanks for your PR!

That's not an appropriate fix. The message is correct but we have to check somewhere else if the parsing goes wrong and throw another error then.

Issue #967, opened 2025-05-05
Support OpenAPI 3.1 Annotated Enums


markrzen:
- https://swagger.io/specification/#annotated-enumerations
- https://www.speakeasy.com/openapi/schemas/enums#enums-with-names-and-values

It appears you already have support for the `oneOf` keyword.

It may be as simple as supporting the `const` keyword, including `title` and `description`.

markrzen:
I may be missing the support for `const` but I suspect that might be the problem to solve:
https://json-schema.org/understanding-json-schema/reference/const

PR #968, opened 2025-05-05
feat: support const

markrzen:
Resolves: #967

You can add the following to a 3.1.1 schema to see these changes in action in the shipping tracking API.


markrzen:
@mosch I have updated the branch, adding the Shipping Tracking API, for an explicit 3.1.1 example using annotated enumeration in three different ways:
1. single string const
2. multiple string const
3. array of string const

mosch:
Great, thank you 👏

PR #972, opened 2025-05-07
feat: improve schema example generation (const, format, *Of)

markrzen:
After adding support for `const` I noticed the "Example Responses" were missing support for some obvious example data, and defaulting to less useful text.


markrzen:
@mosch Thanks so much. Any idea when the two commits I made will be packaged into a release?

markrzen:
Just saw the dev tag. Thanks so much!

PR #989, opened 2025-05-14
Navigation refactoring

mosch:
- We use `path` instead of `navigationId` & `id`

**Todo**
- [x] Rename types from `sidebar` to `navigation`
- [ ] Docs
- [ ] Migration path
- [ ] index file handling

Includes the changes from #1111, however with updated Zod v4


docs/pages/configuration/api-catalog.md
```diff
@@ -13,7 +13,7 @@ The first step to enable the API Catalog, you need to add a `catalogs` object to
 const config = {
   // ...
   catalogs: {
```
joshtwist, line 15: I personally think we should always show an array [] and actually, require it. Making the array optional is just making for a dirty conceptual model imo.

mosch: The scope of this change is only around navigation.

PR #1034, opened 2025-05-21
Fix React key issues in footer

dan-lee:


lcampos (commented):
Reviewed and approved

PR #1039, opened 2025-05-21
Add metadata to schema list page

AdrianMachado:
## Summary

We generated a [schemas page](https://zuplo.com/docs/api/~schemas) but forgot to add metadata toit. This PR adds metadata to the schema list page to improve SEO and accessibility


dan-lee (approved):
Great catch, thanks for the fix! 👏

PR #1046, opened 2025-05-23
Update templates

dan-lee:
ZUP-4542
ZUP-4543
ZUP-4544


website/public/zudoku-logo-full-dark.svg
dan-lee, line 1: These are also unused and outdated

Issue #1031, opened 2025-05-20
Playground sends call twice


rlkasparsolutions:
I just noticed some weird behavior. For some reason my local playground sends a call twice in a row. But the response pane only shows the result of the last call. In our case this results in 201 followed by a 409:

![Image](https://github.com/user-attachments/assets/52ef5688-a43a-427c-960c-c544ca421d92)

dan-lee:
Should be fixed in [v0.46.2](https://github.com/zuplo/zudoku/releases/tag/v0.46.2)!

Issue #1030, opened 2025-05-20
Playground POST body use example button not working


ghost:
As the title says. While using the playground for a POST request I noticed the button does not work at all. I first thought it was specific for some content types (see #1025) but it seems to be in general.

![Image](https://github.com/user-attachments/assets/c382560c-316a-41c4-829d-4d3f042b897e)

dan-lee:
Should be fixed in [v0.46.2](https://github.com/zuplo/zudoku/releases/tag/v0.46.2)!

ghost:
No, it has not. I still cannot use that button. Please re-open this issue.

![Image](https://github.com/user-attachments/assets/b59dc140-11ec-4352-b181-29e58eb9c506).

dan-lee:
Can you provide a reproducible example? In our test example project it looks like it's working: https://www.cosmocargo.dev/api-shipments/rates-and-billing#calculate-shipping-rates

victormer:
> Can you provide a reproducible example? In our test example project it looks like it's working: https://www.cosmocargo.dev/api-shipments/rates-and-billing#calculate-shipping-rates

Hey! Could you please share the API’s openapi.yaml file? I suspect the examples might not be built correctly, and seeing the spec would help clarify. Thanks!

dan-lee:
@victormer You can see all the schemas of this example in this monorepo: https://github.com/zuplo/zudoku/blob/ab6f2529cb90245a3297e9c1e9b5e1e9ebc60916/examples/cosmo-cargo/schema/shipments.json#L1170-L1191

Or [in the directory](https://github.com/zuplo/zudoku/tree/main/examples/cosmo-cargo/schema) directly

victormer:
oh, so 

> [@victormer](https://github.com/victormer) You can see all the schemas of this example in this monorepo:
> 
> [zudoku/examples/cosmo-cargo/schema/shipments.json](https://github.com/zuplo/zudoku/blob/ab6f2529cb90245a3297e9c1e9b5e1e9ebc60916/examples/cosmo-cargo/schema/shipments.json#L1170-L1191)
> 
> Lines 1170 to 1191 in [ab6f252](/zuplo/zudoku/commit/ab6f2529cb90245a3297e9c1e9b5e1e9ebc60916)
> 
>        "examples": { 
>          "basic": { 
>            "summary": "Basic service level", 
>            "value": { 
>              "serviceLevel": "STANDARD" 
>            } 
>          }, 
>          "insured": { 
[...]

Issue #1023, opened 2025-05-20
Content binary pdf not working


rlkasparsolutions:
We have an endpoint where a PDF file is being fetched. We expected the PDF file to return as either a PDF viewer in the playground or an URL like `Binary response. You can download the file manually at *insert link*`. Instead Zudoku tries to return the entire binary contents as string resulting in a HTTP response the size of over 4000000 bytes long, which zudoko will try to render on the page as a string causing the entire browser to freeze and the PC to crash. 

Also, in the `Responses` section on the page it also just says it returns a string. This should be saying "returning a PDF" or something the like. I attached two screenshot. The second is the `responses` section of this particular call.
![Image](https://github.com/user-attachments/assets/14e96e6f-a7fe-4db4-94ce-fc4ce46a6e72)

![Image](https://github.com/user-attachments/assets/cad91e41-01b8-4f1f-8c9f-5b0455dded8d)

dan-lee:
Should be fixed in [v0.46.2](https://github.com/zuplo/zudoku/releases/tag/v0.46.2)!

luanmm:
It would be nice if "Playground" could be extensible. I have a requisite to make a "PDF preview", but can't do it unless the library is patched to make it internally.

Is this something that is being considered for the future? Would a PR be accepted for it, maybe?

Issue #1020, opened 2025-05-20
Size NaN undefinedB when 401


rlkasparsolutions:
We have noticed that when our API returns a 401 and no body Zudoku is unable to determine the body size it seems when the content-length is either 0 or no body was present.

![Image](https://github.com/user-attachments/assets/df211e1f-1799-4495-9cfc-a6d55aa8f0d5)

![Image](https://github.com/user-attachments/assets/1e598cdd-ef1e-4237-b755-2af7d4e8c244)

dan-lee:
Should be fixed in [v0.46.2](https://github.com/zuplo/zudoku/releases/tag/v0.46.2)!

PR #1074, opened 2025-05-28
Playground UI

mosch:
Fixes:
- #1031
- #1023
- #1020
- #1030


packages/zudoku/src/lib/plugins/openapi/playground/Playground.tsx
```diff
@@ -384,18 +370,62 @@ export const Playground = ({
...
+                <div className="items-center px-2 font-mono text-xs break-all leading-6 relative h-full w-full">
+                  <div className="h-full py-1.5">
+                    {serverSelect}
+                    {path}
+                    {urlQueryParams.length > 0 ? "?" : ""}
+                    {urlQueryParams}
+                  </div>
+                  <motion.div
+                    key={mutationId}
+                    className="h-[1px] bg-primary absolute left-0 -bottom-0 z-10"
```
dan-lee: We could've used NProgress here like we do in  https://github.com/zuplo/zudoku/blob/main/packages/zudoku/src/lib/components/TopNavigation.tsx

packages/zudoku/src/lib/plugins/openapi/playground/Playground.tsx
```diff
@@ -195,8 +195,13 @@ export const Playground = ({
     }
   }, [latestSetRememberedIdentity, formState.identity]);
 
+  const [mutationId, setMutationId] = useState(crypto.randomUUID());
+
```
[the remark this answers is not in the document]
dan-lee: Not entirely sure about the use of `crypto` here, wouldn't a `Date.now()` suffice?

packages/zudoku/src/lib/plugins/openapi/playground/Playground.tsx
```diff
@@ -405,142 +435,35 @@ export const Playground = ({
...
+                    setValue={(value) => setValue("identity", value)}
+                  />
                 </div>
-              </TabsContent>
-            </Tabs>
+              </div>
+            )}
+
+            {pathParams.length > 0 && (
+              <div className="flex flex-col gap-2   ">
```
dan-lee:
```suggestion
              <div className="flex flex-col gap-2">
```

packages/zudoku/src/lib/plugins/openapi/playground/Playground.tsx
```diff
@@ -405,142 +435,35 @@ export const Playground = ({
...
+                <span className="font-semibold">Path Parameters</span>
+                <PathParams url={url} control={control} />
+              </div>
+            )}
+
+            <div className="flex flex-col gap-2 ">
```
dan-lee:
```suggestion
            <div className="flex flex-col gap-2">
```

packages/zudoku/src/lib/ui/Dialog.tsx
```diff
@@ -42,10 +41,6 @@ const DialogContent = React.forwardRef<
       {...props}
     >
       {children}
-      <DialogPrimitive.Close className="absolute right-4 top-4 rounded-xs opacity-70 ring-offset-background transition-opacity hover:opacity-100 focus:outline-hidden focus:ring-2 focus:ring-ring focus:ring-offset-2 disabled:pointer-events-none data-[state=open]:bg-accent data-[state=open]:text-muted-foreground">
-        <X className="h-4 w-4" />
-        <span className="sr-only">Close</span>
-      </DialogPrimitive.Close>
```
dan-lee, line 48: Maybe we should do it like shadcn's dialog component does now in the current version with a prop `showCloseButton` instead of removing it entirely. See https://ui.shadcn.com/docs/components/dialog#installation

dan-lee (commented):
Great work! Left some comments.

Also the code block of the response has this weird issue of overlapping text with the icon:

CleanShot 2025-06-03 at 11 33 41

dan-lee (approved):

PR #1081, opened 2025-06-02
Bring back `rehype-raw`

### LINKED ISSUE #1078: Embedded HTML Breaking Changes for OpenAPI docs in 0.45.1
Hi, we recently upgraded to version **0.45.1** from **0.44.1** and noticed breaking changes for embedded HTML in OpenAPI docs.

**Before**:

Image

**After**:

Image

It looks like this breaking change was caused by this PR https://github.com/zuplo/zudoku/pull/961.

**Suggested fix**: Add configuration option to preserve HTML rendering in OpenAPI docs while using Shiki for code blocks.

Or

**Conditional Shiki Application**
The proper fix would be to modify how Shiki is applied. Based on the standard approach with rehype-pretty-code [Rehype Pretty Code | Rehype Pretty](https://rehype-pretty.pages.dev/), Shiki should only process code blocks:

dan-lee:
Fixes #1078

### DIFF
--- a/packages/zudoku/package.json
+++ b/packages/zudoku/package.json
@@ -256,7 +256,6 @@
     "react-markdown": "10.1.0",
     "react-router": "7.5.2",
     "rehype-mdx-import-media": "1.2.0",
-    "rehype-pretty-code": "0.14.1",
     "rehype-raw": "7.0.0",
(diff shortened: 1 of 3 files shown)


eddiedozier:
Thank you! This fixed the issue!

PR #1118, opened 2025-06-04
Update Zuplo example

mosch:


examples/with-zuplo/docs/pages/banner.jpg
dan-lee, line 1: Didn't you want to replace it with an SVG?

dan-lee (approved):

PR #1155, opened 2025-06-09
Support pre-defined examples for path and query params in playground

deividfortuna:
Support pre-defined examples for path and query params in playground following OpenApi spec

### DIFF
--- a/examples/cosmo-cargo/schema/shipments.json
+++ b/examples/cosmo-cargo/schema/shipments.json
@@ -1014,6 +1014,7 @@
             "name": "holdDuration",
             "in": "query",
             "description": "Duration to hold the shipment",
+            "example": "2017-07-21T17:32:28Z",
             "schema": {
               "type": "string",
               "format": "date-time"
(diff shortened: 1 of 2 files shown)


mosch:
@deividfortuna thanks for bringing this up in. We discussed this change and we think 
it's better: 

- **to have a little "info" icon that shows the exmaple value** 

_or_ 

- **use the example as the placeholder of the input field.**

The reason is that we believe the example value is usually not valid; it serves to illustrate what a value might look like.


For your use case, could you use the defaultValue to retrieve the values in the Playground?

deividfortuna:
> @deividfortuna thanks for bringing this up in. We discussed this change and we think it's better:
> 
> * **to have a little "info" icon that shows the exmaple value**
> 
> _or_
> 
> * **use the example as the placeholder of the input field.**
> 
> The reason is that we believe the example value is usually not valid; it serves to illustrate what a value might look like.
> 
> For your use case, could you use the defaultValue to retrieve the values in the Playground?

Thanks @mosch, I think you have a point, I'll have another look later today

deividfortuna:
default value works, thank you

PR #1157, opened 2025-06-09
Require ZUPLO_PUBLIC_ env vars

ntotten:
The Zuplo mode env vars are supposed to be `ZUPLO_PUBLIC_`, not just `ZUPLO_`


mosch:
This might break the usage of a bunch of `ZUPLO_` variables like Looks `ZUPLO_ENVIRONMENT_TYPE`, `ZUPLO_BUILD_ID` and `ZUPLO_BUILD_CONFIG`.

We could consider allow-listing them.

PR #1230, opened 2025-06-18
UI fixes playground

mosch:


dan-lee (approved):

packages/zudoku/src/lib/plugins/openapi/playground/result-panel/ResponseTab.tsx
```diff
@@ -82,6 +83,11 @@ const sortHeadersByRelevance = (
 
 const MAX_HEADERS_TO_SHOW = 3;
 
+const Row = createVariantComponent(
+  "div",
+  "grid-cols-subgrid grid border-b col-span-full px-4 py-1.5 font-mono text-xs min-h-10 items-center",
```
dan-lee: I don't think it should be `items-center`, looks really odd for multiple lines. Other tools like Chrome or HTTPie don't align it center as well

PR #1233, opened 2025-06-18
Add Description support in MDX

AdrianMachado:
## Summary

We done goofed. We documented [this property](https://zudoku.dev/docs/markdown/frontmatter#description) but lost support somewhere along the way.

I am also fixing a bug with the `title` property not being respected - it should override the ToC entry if explicitly set IMO

### DIFF
--- a/packages/zudoku/src/lib/plugins/markdown/MdxPage.tsx
+++ b/packages/zudoku/src/lib/plugins/markdown/MdxPage.tsx
@@ -53,10 +53,11 @@ export const MdxPage = ({
   const categoryTitle = useCurrentItem()?.categoryLabel;
 
   const title = frontmatter.title;
+  const description = frontmatter.description ?? excerpt;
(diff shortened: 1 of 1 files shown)


mosch (approved):

packages/zudoku/src/lib/plugins/markdown/MdxPage.tsx
```diff
@@ -110,7 +111,7 @@ export const MdxPage = ({
     >
       <Helmet>
         <title>{pageTitle}</title>
-        {excerpt && <meta name="description" content={excerpt} />}
```
mosch, line 113: We are changing the behaviour here - we should at least fall back to the old behaviour?

AdrianMachado: Shouldn't line 56 handle that @mosch ? If there is no description it uses the excerpt instead. If neither exist, nothing shows up

PR #1241, opened 2025-06-19
Add zuplo template

ntotten:
Adds zuplo template in create app. This template will be pulled into the zuplo create project when zudoku is released.


vazexqi (commented):
Reviewed and approved

PR #1262, opened 2025-06-25
Fix status page

dan-lee:


packages/zudoku/src/app/main.tsx
```diff
@@ -125,6 +116,7 @@ export const getRoutesByConfig = (config: ZudokuConfig): RouteObject[] => {
           <Layout />
         </Zudoku>
       ),
+      hydrateFallbackElement: <div>Loading...</div>,
```
mosch, line 119: Hah, maybe we can add something nicer here?

PR #1273, opened 2025-06-27
Fix link

AdrianMachado:
## Summary

This link breaks when it gets ported into Zuplo docs because of the way we map paths. I will fix that mapping function later, for now I am making this URL consistent with how we reference other pages.


lcampos (commented):
Reviewed and approved

PR #1276, opened 2025-06-29
Fix plugin extensions in package.json

jimbz:
The missing extensions lead to package resolution errors when using the affected plugins. For example the supabase example did not build before this fix.


dan-lee:
Good finding! Out of curiosity how did you stumble over this? For the published package we use the `publishConfig.exports`:

https://github.com/zuplo/zudoku/blob/446858f2ea63ae7b914a395089c36cff2f4620f9/packages/zudoku/package.json#L53-L162

jimbz:
Thank you for the quick turnaround. To be honest I don't really know what I am doing :). I got this error when following the Supabase integration documentation https://zudoku.dev/docs/configuration/authentication#supabase:
```
[vite]: Rollup failed to resolve import "zudoku/auth/supabase" from "virtual:zudoku-auth".
```

So I cloned the repository and tried to build the supabase example and noticed this discrepancy. I haven't figured out how to test if this PR fixes my initial issue yet.

dan-lee (approved):

dan-lee:
Ah, I see! 😅 Great that you found the reason and fix for it. Thanks for the contribution!

PR #1297, opened 2025-07-06
Fix the Supabase plugin

jimbz:
I am not sure if this is the right fix but the following changes seem to get the Supabase plugin to work:

 * Add the supabase plugin to the vite.config.ts, like the other plugins
 * Make sure the supabase dependency (a peer dependency) is considered external and not bundled, which would cause: `ReferenceError: require is not defined in ES module scope, you can use import instead`.


dan-lee (approved):
Great finding @jimbz, thanks for your contribution. Looks good to me!

PR #1317, opened 2025-07-10
Update GTM docs & fix button

AdrianMachado:
## Summary

Addressing 2 things from working with Accuweather

1. Their Google Tag Manager was not capturing any page navigations which hampered their ability to understand user behavior. This is because of React Router, so i added docs on how to write an easy plugin to log to GTM when page changes
2. We nested a div inside of a button, which I recall is generally not advisable. Maybe a `span` is better but I figure having a div when the button is disabled and content is invisible can't hurt

### DIFF
--- a/packages/zudoku/src/lib/ui/ActionButton.tsx
+++ b/packages/zudoku/src/lib/ui/ActionButton.tsx
@@ -19,7 +19,9 @@ export const ActionButton = forwardRef<HTMLButtonElement, ActionButtonProps>(
             <Spinner />
           </div>
         )}
-        <span className={cn("block", isPending && "invisible")}>{children}</span>
+        <span className={cn("block", isPending && "invisible")}>
+          {children}
+        </span>
       </Button>
     );
   },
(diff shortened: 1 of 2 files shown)


packages/zudoku/src/lib/ui/ActionButton.tsx
```diff
@@ -19,7 +19,7 @@ export const ActionButton = forwardRef<HTMLButtonElement, ActionButtonProps>(
             <Spinner />
           </div>
         )}
-        <div className={cn(isPending && "invisible")}>{children}</div>
+        {isPending ? <div className="invisible">{children}</div> : children}
```
dan-lee: What if we used a `span` as `block` element? (similar what you've suggested)

```suggestion
        <span className={cn("block", isPending && "invisible")}>{children}</span>
```

AdrianMachado: Yeah, I believe this works too - but wanted to defer decision to y'all

dan-lee (approved):

PR #1333, opened 2025-07-14
Fix category document items

dan-lee:
Follow up for #1318

We also need to take category items with links to a document into consideration


mosch:
Reviewed & Approved

PR #1341, opened 2025-07-16
Allow to serve all doc files

dan-lee:


packages/zudoku/src/config/validators/validate.ts
```diff
@@ -228,6 +228,13 @@ export const DocsConfigSchema = z.object({
     .union([z.string(), z.array(z.string())])
     .transform((val) => (typeof val === "string" ? [val] : val))
     .default([DEFAULT_DOCS_FILES]),
+  serveAllFiles: z
```
mosch: I don't like the name of this option, `enableFileRouter` or `enableMarkdownRoutes` 🤷‍♂️

packages/zudoku/src/vite/plugin-docs.ts
```diff
@@ -93,25 +91,24 @@ const viteDocsPlugin = (): Plugin => {
...
+          const customPath = ensureLeadingSlash(doc.path);
+          globbedDocuments[customPath] = importPath;
+          delete globbedDocuments[fileRoutePath];
         });
       }
 
       code.push(
         `const fileImports = {`,
-        ...Object.entries(navigationFileImports).map(
+        ...Object.entries(globbedDocuments).map(
```
dan-lee, line 111: Reverted here

PR #1343, opened 2025-07-16
Added a lint step to check admonition formatting

ntotten:
- Fixes admonition new lines in docs
- Adds a script and github action check to ensure future admonitions are formatted correctly


dan-lee:
Reviewed and approved

PR #1378, opened 2025-07-23
Work on Landingpage

mosch:
Adressing a feedback round with Yoh and Martyn

### DIFF
--- a/docs/src/components/Preview.tsx
+++ b/docs/src/components/Preview.tsx
@@ -11,12 +11,14 @@ const tags = [
 ];
 
 const Hero = () => {
-  const tag = useMemo(() => {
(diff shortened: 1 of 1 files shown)


docs/src/LandingPage.tsx
```diff
@@ -472,24 +497,24 @@ const LandingPage = () => {
               </BentoImage>
               <BentoDescription
                 title="Ready to use Components"
-                description="Integrates easily with easy extensible with own plugins community and core plugins"
+                description="ICreate the developer experience you've always dreamed of with a full suite of reusable components (or create your own)."
```
dan-lee:
```suggestion
                description="Create the developer experience you've always dreamed of with a full suite of reusable components (or create your own)."
```

PR #1419, opened 2025-07-30
Remove changelog trigger

ntotten:
I am changing the changelog to be a weekly change instead of running on every minor release.


vazexqi (commented):
Reviewed and approved

PR #1431, opened 2025-07-31
Treat `.md` as plain Markdown

### LINKED ISSUE #1415: `*.md` files are interpreted as `*.mdx` files
It seems that plain Markdown `*.md` files are being interpreted as `*.mdx` files, which causes errors on seemingly valid Markdown, such as curly braces (`{foo}`) being used to interpret `foo`, which won't work as the plain Markdown file hasn't been set-up for this use. This also doesn't seem to be an issue from [GitHub Flavored Markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) (GFM) as this works just fine on GitHub.

The Zudoku documentation also mentions the following in [Getting Started](https://zudoku.dev/docs/markdown/mdx#getting-started):

> To use MDX in your documentation, simply create files with the .mdx extension instead of .md. These files work exactly like regular markdown files but with all MDX features unlocked - you can write normal markdown content and add JSX components whenever needed.

Which seems to suggested that this is an unintended side-effect, and that `.md` files should not be parsed as `.mdx` files.

---

For example, this file (`introduction.md`) is invalid

```md title=introduction.md
This is a descriptive introduction about {foo}.
```

However, this file (also `introduction.md`) is valid, but needs additional escapes to create cruly brace literals. Which shouldn't be necessary for a plain Markdown file.

```md title=introduction.md
This is descriptive introduction about \{foo\}.
```

Is this intended, and how do we prevent plain Markdown files being treated as MDX files?

dan-lee:
Fixes #1415

⚠️  Kind of a breaking change if one relied on `.md` files being treated as MDX.

### DIFF
--- a/packages/zudoku/src/vite/plugin-mdx.ts
+++ b/packages/zudoku/src/vite/plugin-mdx.ts
@@ -85,9 +85,8 @@ const viteMdxPlugin = (): Plugin => {
         config.__meta.mode === "internal" || config.__meta.mode === "standalone"
           ? "@mdx-js/react"
           : "zudoku/components",
-      // Treat .md files as MDX
-      mdxExtensions: [".md", ".mdx"],
-      format: "mdx",
+      mdxExtensions: [".mdx"],
+      format: "detect",
       remarkPlugins: [
         remarkStaticGeneration,
         [remarkInjectFilepath, config.__meta.rootDir],
--- a/examples/cosmo-cargo/pages/global.md
+++ b/examples/cosmo-cargo/pages/global.md
@@ -1,0 +1,70 @@
+---
+sidebar_label: Global Shipping
+sidebar_icon: globe
+---
+
+# Global Shipping
+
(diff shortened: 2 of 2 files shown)


packages/zudoku/src/vite/plugin-mdx.ts
```diff
@@ -85,9 +85,8 @@ const viteMdxPlugin = (): Plugin => {
         config.__meta.mode === "internal" || config.__meta.mode === "standalone"
           ? "@mdx-js/react"
           : "zudoku/components",
-      // Treat .md files as MDX
-      mdxExtensions: [".md", ".mdx"],
-      format: "mdx",
+      mdxExtensions: [".mdx"],
+      format: "detect",
```
dan-lee, line 89: Main change

mosch (approved):

PR #1437, opened 2025-08-01
Use `x-displayName` in category title

dan-lee:


packages/zudoku/src/lib/plugins/openapi/OperationList.tsx
```diff
@@ -245,7 +244,7 @@ export const OperationList = ({
                 registerNavigationAnchor
                 className="mb-0"
               >
-                {schema.tag.name ?? "Documentation"}
+                {tagTitle}
```
dan-lee, line 247: Main change

PR #1439, opened 2025-08-03
FIX: mixed localized dates and prefixes

dbrrt:
The date displayed should be consistent with the displayed language (here english) and shouldn't fallback to browser language.

Capture d’écran du 2025-08-03 03-10-00

<br />

Capture d’écran du 2025-08-03 03-24-44


dan-lee:
@dbrrt  Thanks for your contribution! 🚀

PR #1469, opened 2025-08-14
Fix `anyOf` variants

dan-lee:
Inherit parent properties in `anyOf`/`oneOf`. Also use labels in decision table to switch between the variants, no need for the duplicated the TabsList

### DIFF
--- a/packages/zudoku/src/lib/plugins/openapi/schema/UnionView.tsx
+++ b/packages/zudoku/src/lib/plugins/openapi/schema/UnionView.tsx
@@ -1,4 +1,3 @@
-import { ArrowDownIcon } from "lucide-react";
 import { useState } from "react";
 import type { SchemaObject } from "../../../oas/parser/index.js";
 import { Badge } from "../../../ui/Badge.js";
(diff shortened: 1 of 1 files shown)


packages/zudoku/src/lib/plugins/openapi/schema/UnionView.tsx
```diff
@@ -94,30 +123,22 @@ export const UnionView = ({ schema }: { schema: SchemaObject }) => {
...
-        <DecisionTable variants={variants} schema={schema} />
+        <DecisionTable
+          variants={variants}
+          schema={schema}
+          selectedVariant={selectedVariant}
+          onSelectVariant={setSelectedVariant}
+        />
+
+        <div className="flex justify-center items-center gap-2">
+          <ArrowDownIcon size={14} />
```
lcampos: Are these 4 arrows correct ?

lcampos (approved):

packages/zudoku/src/lib/plugins/openapi/schema/UnionView.tsx
dan-lee: Yeah good catch, had this in to test stuff. should probably look a bit different

PR #1521, opened 2025-09-03
Allow to configure prerender workers

dan-lee:


vazexqi (commented):
Reviewed and approved.

PR #1530, opened 2025-09-05
Fix remark-directive parsing issues

dan-lee:
After upgrading to `remark-directive@4.0.0`, time formats in MDX content (like `9:30:00`) are now being parsed as directives, and via `remark-directive-rehype` resulted in invalid HTML elements/React components (e.g. `<30>`) that caused React rendering errors.

As workaround I'm downgrading for now


martyndavies (approved):
LGTM

PR #1558, opened 2025-09-17
Auth0 prompt parameter omitted instead of being set to "login".

rikard-swahn:
prompt=login forces reauthentication, so that password needs to be entered every time, even when logged in.

prompt parameter omitted means Auth0 will redirect the user to the login page if no valid session exists. If a valid session is present, the user will be silently authenticated.

The best solution would be to have prompt configurable. But until then, it seems like omitting parameter is a better default than "login". This change might count as breaking though.

### DIFF
--- a/packages/zudoku/src/lib/authentication/providers/auth0.tsx
+++ b/packages/zudoku/src/lib/authentication/providers/auth0.tsx
@@ -25,7 +25,6 @@ class Auth0AuthenticationProvider
     url: URL,
     { isSignUp }: { isSignUp: boolean },
   ) => {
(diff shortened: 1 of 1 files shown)


mosch:
Thanks for that PR, and I think your intention is right. Let me do some testing to see if it might break something.

rikard-swahn:
> Thanks for that PR, and I think your intention is right. Let me do some testing to see if it might break something.

Hello. I just wanted to check if you have had time to look more at this?

dan-lee:
Hey thanks for the PR and the reminder @rikard-swahn. We went ahead and added it as an option in #1613.

PR #1572, opened 2025-09-29
feat: add `expandApiInformation` to auto-expand API information

atissedrescaleway:
## Description
Adds a new boolean option `expandApiInformation` in API options. When `true`, the “API information” collapsible renders expanded by default. Default is `false` (no behavior change).

**Why**
Some teams want the top-level API details visible on first load without user interaction.

## Config usage

```
// zudoku.config.ts

export const config = {
  defaults: {
    apis: {
      expandApiInformation: false, // default (optional)
    },
  },
  apis: {
    type: "file",
    input: "./openapi.json",
    path: "/api",
    options: {
      expandApiInformation: true, // opt-in
    },
  },
};
```


atissedrescaleway:
No problem at all, thanks for taking the time to review !

PR #1591, opened 2025-10-06
Corrected quick-start URL

LeonMelis:
Following the URL for the quick-start documentation from the README.md results in a 404.

The correct URL seems to be: https://zudoku.dev/docs/quickstart

I would recommend the Zuplo website administrators to also create a permanent (301) redirect from `/docs/app-quickstart` to `/docs/quickstart` for users of older branches/versions.


mosch:
Thank you! 👏

PR #1612, opened 2025-10-17
Invalidate auth state on logout

mosch:


packages/zudoku/src/lib/components/context/ZudokuContext.ts
```diff
@@ -23,6 +24,14 @@ export const useZudoku = () => {
 
 export const useApiIdentities = () => {
   const { getApiIdentities } = useZudoku();
+  const { isAuthenticated } = useAuthState();
+  const { invalidateCache } = useCache();
+
+  useEffect(() => {
+    if (isAuthenticated) {
```
dan-lee:
```suggestion
    if (!isAuthenticated) {
```

Shouldn't this be negated?

PR #1614, opened 2025-10-19
Add support for additional programming languages

dbrrt:
Add support for c, commonlisp, c++, dart, elixir, ocaml, powershell, scala, zig

### DIFF
--- a/docs/pages/docs/components/syntax-highlight.mdx
+++ b/docs/pages/docs/components/syntax-highlight.mdx
@@ -407,3 +407,197 @@ struct User: Codable {
   </modules>
 </configuration>
 ```
+
+### C
+
+```c title="hello.c"
+#include <stdio.h>
+#include <stdlib.h>
+
+typedef struct {
+    int id;
+    char name[50];
+} User;
+
+int main() {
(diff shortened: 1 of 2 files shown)


biome.jsonc
```diff
@@ -37,8 +37,7 @@
       "recommended": true,
       "suspicious": {
         "noConsole": "error",
-        "noConfusingVoidType": "off",
```
dan-lee, line 40: The lint check now fails because of this change. Any reason why this was included?

dbrrt: I can reenable it, it wasn't building on my side with this flag

dan-lee: Hm that's odd, do you still have the output?

dbrrt: I don't but I can try to reproduce it out of business hours

dan-lee:
Thanks again for your PR! I've added these languages to the docs as well and optimized them further to work better in Zudoku context.

You can see the preview here: https://zudoku-docs-git-fork-dbrrt-add-zig-scala-dart-ocaml-c-cp-a99576.zuplosite.com/docs/components/syntax-highlight#supported-languages

PR #1662, opened 2025-10-30
Add opt-in Mermaid diagram support for MDX files

ankitpandey2708:
Introduces comprehensive Mermaid diagram rendering for Zudoku with a flexible, zero-dependency approach.

### **Key Features**

- **Opt-in design**: No hard dependency on `mermaid` package - users install only if needed

- **Client-side rendering** (recommended): Lightweight zero-dependency rehype plugin transforms code blocks for browser rendering

- **Server-side rendering** (optional): Support for `rehype-mermaid` when SSG is required

### DIFF
--- a/packages/zudoku/src/lib/plugins/openapi/playground/result-panel/ResponseTab.tsx
+++ b/packages/zudoku/src/lib/plugins/openapi/playground/result-panel/ResponseTab.tsx
@@ -1,7 +1,5 @@
 import { useQuery } from "@tanstack/react-query";
 import {
-  CheckIcon,
-  CopyIcon,
   CornerDownLeftIcon,
   CornerDownRightIcon,
   DownloadIcon,
@@ -9,7 +7,7 @@ import {
   EyeOffIcon,
   PlusCircleIcon,
 } from "lucide-react";
-import { useRef, useState } from "react";
+import { useState } from "react";
 import { Button } from "zudoku/ui/Button.js";
 import {
   Collapsible,
@@ -27,7 +25,6 @@ import {
 import { cn } from "../../../../util/cn.js";
 import createVariantComponent from "../../../../util/createVariantComponent.js";
 import { humanFileSize } from "../../../../util/humanFileSize.js";
-import { useCopyToClipboard } from "../../../../util/useCopyToClipboard.js";
 import {
   CollapsibleHeader,
   CollapsibleHeaderTrigger,
@@ -143,8 +140,6 @@ export const ResponseTab = ({
   fileName?: string;
   blob?: Blob;
 }) => {
-  const [isCopied, copyToClipboard] = useCopyToClipboard();
-  const ref = useRef<HTMLDivElement>(null);
   const detectedLanguage = detectLanguage(headers);
   const jsonContent = tryParseJson(body);
   const beautifiedBody = jsonContent || body;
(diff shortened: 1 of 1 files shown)


ankitpandey2708:
@dan-lee @mosch  Please look into it.

https://github.com/user-attachments/assets/7689226b-7d32-4233-b448-85cd0fa1351a

ankitpandey2708:
@dan-lee 

# 1. Why We Need to Modify `shiki.ts`

The core issue is that **Shiki completely replaces `<pre>` and `<code>` elements** during its processing (documented in [shikijs/shiki#629](https://github.com/shikijs/shiki/issues/629)).

When users write:

```mermaid
graph LR
  A --> B
```

The markdown processor creates:

```html
<pre><code class="language-mermaid">graph LR\n  A --> B</code></pre>
```

But after Shiki processes it, the `language-mermaid` class is lost because:

- Shiki doesn’t recognize `"language-mermaid"` as a bundled language.
- It replaces the entire element with newly generated HTML.
- Any properties/classes from the original are discarded.

Without modifying `shiki.ts`, our downstream `rehype-mermaid-client` plugin has **no way to detect** which blocks are mermaid diagrams.

---

# 2. If Issue [#629](https://github.com/shikijs/shiki/issues/629) Was Solved

If Shiki preserved data and properties from original HAST nodes, our `shiki.ts` wouldn’t need any modifications. The file would look exactly as it does today minus the mermaid code

`language-mermaid` class would naturally survive through Shiki, and our client plugin would just work.

---

# 3. Current Solution (After Research)

After trying multiple approaches (CSS classes, data attributes, `langAlias`, parent marking), we discovered that **Shiki preserves meta strings** for transformers — this became the key to our solution.

### Additions to `shiki.ts` (~40 lines)

#### 1. Preprocessor: Mark mermaid blocks using meta string

Shiki preserves meta strings, so we use them to flag mermaid code blocks.

```ts
const rehypeMermaidPreprocessor = () => (tree: Root) => {
  visit(tree, "element", (node, _index, _parent) => {
    if (node.tagName !== "code") return;

    const classes = Array.isArray(node.properties.className)
      ? node.properties.className
      : [];
    const isMermaid = classes.some(
      (c) =>
        typeof c === "string" &&
        (c === "language-mermaid" || c === "lang-mermaid" || c === "mermaid")
    );

    if (isMermaid) {
      node.data = node.data || {};
      node.data.meta = "zudoku-mermaid"; // Meta strings survive Shiki
    }
  });
};
```

#### 2. Transformer: Runs *during* Shiki to mark output

This transformer checks for the `zudoku-mermaid` marker and ensures it’s preserved in Shiki’s output.

```ts
const transformerMermaid = (): ShikiTransformer => ({
  name: "mermaid-marker",
  code(node) {
    const meta = this.options.meta?.__raw || "";

    if (meta.includes("zudoku-mermaid")) {
      node.properties["data-mermaid"] = "true";
      const classes = Array.isArray(node.properties.className)
        ? node.properties.className
        : [];
      if (!classes.includes("language-mermaid")) {
        classes.push("language-mermaid");
      }
      node.properties.className = classes;
    }
  },
});
```

### Why This Works

1. **Preprocessor** detects `language-mermaid` → sets `node.data.meta = "zudoku-mermaid"`.
2. Shiki **preserves meta strings**, which transformers can access.
3. **Transformer** runs *during* Shiki processing → sees `this.options.meta.__raw` contains our marker.
4. **Transformer** adds `data-mermaid` and class markers to output → these survive since they’re part of Shiki’s generated HTML.

This pattern — using transformers with meta strings — is **the recommended way** to customize Shiki output.

---

ankitpandey2708:
@dan-lee any update?

dan-lee:
Thanks for the effort on this PR! However, I have concerns about the complexity and maintenance burden this introduces to the core framework.

## Main Issues

1. **Two unrelated features** - The Mermaid support and the copy button for API responses should be separate PRs
2. **High complexity** - Custom Shiki transformers with workarounds, 4 different rendering strategies, optional dependency management across multiple files, etc.
3. **Large maintenance surface** - This touches critical areas (Shiki integration) and creates ongoing maintenance obligations for the core framework

## Alternative Approach

After reviewing your PR, I explored whether we could solve the core problem (plugin ordering) without adding complexity to Zudoku core. I've tried an alternative implementation that demonstrates how users can add Mermaid support using standard `rehype-mermaid`:

**The Problem You Identified Was Real**: Plugin ordering - `rehype-mermaid` needs to run BEFORE Shiki, but users couldn't control that.

**Alternative Solution**: Function-based plugin configuration that gives users control over plugin ordering:

```typescript
// zudoku.build.ts
import rehypeMermaid from "rehype-mermaid";

export default {
  rehypePlugins: (plugins) => [
    // Insert rehype-mermaid BEFORE Shiki and other plugins
    [rehypeMermaid, { strategy: "inline-svg" }],
    ...plugins,
  ],
};
```

This approach:

- ✅ Works with standard `rehype-mermaid` (no Zudoku-specific code needed)
- ✅ Solves the plugin ordering problem you identified
- ✅ No maintenance burden or complexity added to Zudoku core
- ✅ Users control their own rehype/remark pipeline
- ✅ No custom Shiki transformers or workarounds required

**See the working example**: I've added `examples/mermaid/` to demonstrate this approach in this PR: https://github.com/zuplo/zudoku/pull/1678

## What About Theme Switching?

The one trade-off: server-rendered diagrams are static (one theme). While dynamic theme switching is nice to have, the complexity and maintenance burden it requires (client-side rendering, hydration handling, custom Shiki transformers, optional dependency management) doesn't justify adding it to core. Users can either pick one theme that matches their design, or implement dynamic theme switching as a community plugin if it's essential for their use case.

What are your thoughts on this alternative approach? Does the function-based plugin config solve the problem you were trying to address?

ankitpandey2708:
@dan-lee 

1. have created a separate PR for copy button. https://github.com/zuplo/zudoku/pull/1679
2. function-based plugin config solves the problem . my only concern was if changing the order had any other effects.

PR #1674, opened 2025-11-03
Implemented supabase login with UI

ntotten:
Adds support for supabase auth with custom UI and ability to support many supabase login providers.


packages/zudoku/src/lib/authentication/components/SignIn.tsx
```diff
@@ -8,17 +8,80 @@ import {
...
-    });
-  }, [context.authentication, search]);
+    if (isAuthenticated) {
+      const redirectTo = search.get("redirect") ?? "/";
+      navigate(redirectTo, { replace: true });
+    }
+  }, [isAuthenticated, navigate, search]);
+
+  // Clear OAuth error when leaving the error page
+  useEffect(() => {
```
mosch, line 46: If you just close the page this will not be cleared, but the error state will remain persisted

packages/zudoku/src/lib/authentication/authentication.ts
```diff
@@ -6,6 +6,34 @@ export interface AuthenticationPlugin {
   getAccessToken(): Promise<string>;
   onPageLoad?(): void;
   signRequest(request: Request): Promise<Request>;
+  /**
+   * Optional method to render custom sign-in UI.
+   * If provided along with hasCustomUI=true, this will be rendered instead of redirecting.
+   * Returns a component function that renders the custom UI.
+   */
+  renderSignInUI?(): () => React.JSX.Element | null;
```
mosch, line 14:
```suggestion
  renderSignIn?(): () => React.JSX.Element | null;
```

Lets rename this to avoid the `UI` suffix

packages/zudoku/src/lib/authentication/authentication.ts
```diff
@@ -6,6 +6,34 @@ export interface AuthenticationPlugin {
...
+  /**
+   * Optional method to render custom sign-up UI.
+   * If provided along with hasCustomUI=true, this will be rendered instead of redirecting.
+   * Returns a component function that renders the custom UI.
+   */
+  renderSignUpUI?(): () => React.JSX.Element | null;
```
mosch, line 20:
```suggestion
  renderSignUp?(): () => React.JSX.Element | null;
```

packages/zudoku/src/lib/authentication/authentication.ts
```diff
@@ -6,6 +6,34 @@ export interface AuthenticationPlugin {
...
+  /**
+   * Indicates whether this provider uses custom UI instead of redirects.
+   * If true, the SignIn/SignUp components will render the custom UI.
+   */
+  hasCustomUI?: boolean;
```
mosch, line 25:
```suggestion
```

I suggest we remove this flag and just by having the `render` methods, Zudoku will do the custom UI. 
Otherwise you may set this flag to `true` without providing the methods.

packages/zudoku/src/lib/authentication/state.ts
```diff
@@ -6,6 +6,11 @@ export interface AuthState<ProviderData = unknown> {
   isPending: boolean;
   profile: UserProfile | null;
   providerData: ProviderData | null;
+  oauthError?: {
```
mosch, line 9:
```suggestion
  error: {
```

Let's use this for all authentication types, we want to keep the interface clean of auth-type specific stuff (see `providerData`)

packages/zudoku/src/lib/authentication/hook.ts
```diff
@@ -37,6 +45,15 @@ export const useAuth = () => {
       if (!isAuthEnabled) {
         throw new Error("Authentication is not enabled.");
       }
+
+      // For providers with custom UI, navigate to signup page
+      if (authentication.hasCustomUI) {
```
mosch, line 50:
```suggestion
      if (authentication.renderSignup) {
```

packages/zudoku/src/lib/authentication/hook.ts
```diff
@@ -16,7 +16,15 @@ export const useAuth = () => {
       if (!isAuthEnabled) {
         throw new Error("Authentication is not enabled.");
       }
-      // TODO: Should handle errors/state
+
+      // For providers with custom UI, navigate to signin page
+      if (authentication.hasCustomUI) {
```
mosch, line 21:
```suggestion
      if (authentication.renderSignIn) {
```

packages/zudoku/src/lib/authentication/components/SignUp.tsx
```diff
@@ -7,15 +8,44 @@ import {
...
+  }, [isAuthenticated, navigate]);
+
+  // Check if provider has custom UI
+  if (
+    context.authentication?.hasCustomUI &&
+    context.authentication?.renderSignUpUI
+  ) {
+    const CustomSignUpUI = context.authentication.renderSignUpUI();
+    return (
+      <div className="flex items-center justify-center mt-8">
```
mosch, line 43: This seems quite opinionated. Maybe we should leave the full responsibility for "well looking integration" to the providers instead of adding potential limitations here.

packages/zudoku/src/lib/authentication/components/SignUp.tsx
```diff
@@ -7,15 +8,44 @@ import {
...
 } from "zudoku/ui/Card.js";
+import { useAuthState } from "../../authentication/state.js";
 import { useZudoku } from "../../components/context/ZudokuContext.js";
 
 export const SignUp = () => {
   const context = useZudoku();
+  const navigate = useNavigate();
+  const { isAuthenticated } = useAuthState();
 
+  // Existing redirect-based flow for OAuth providers
```
mosch, line 19:
```suggestion
```

packages/zudoku/src/lib/authentication/components/SignUp.tsx
```diff
@@ -7,15 +8,44 @@ import {
...
   useEffect(() => {
-    void (context.authentication?.signUp() ?? context.authentication?.signIn());
+    // Only trigger redirect if not using custom UI
```
mosch, line 21:
```suggestion
```

packages/zudoku/src/lib/authentication/components/SignUp.tsx
mosch: remove weird comment

packages/zudoku/src/lib/authentication/components/SignUp.tsx
mosch: remove weird comment

packages/zudoku/src/lib/authentication/providers/supabase.tsx
```diff
@@ -90,31 +73,25 @@ class SupabaseAuthenticationProvider
...
-    });
+  renderSignUpUI = () => {
+    // Supabase Auth UI handles both sign in and sign up
+    return this.renderSignInUI();
   };
 
-  signIn = async ({ redirectTo }: { redirectTo?: string }) => {
-    const finalRedirectTo = redirectTo ?? this.redirectToAfterSignIn;
+  // No-op methods since Auth UI handles the interactions
+  signUp = async () => {
```
mosch, line 89: 🚨 Big warning that something is off. If this is not required to do something but is required by the interface, then the something is off.

mosch (changes requested):
Pretty good ideas, but I think we have to solve some of the interface issues here.

If feel we're doing it wrong by making the `signIn` and `signUp` an non-op and adding `render...` Functions to the interface.

Maybe signIn should redirect to the custom pages instead. Let me try to see if that would work....

packages/zudoku/src/lib/components/MobileTopNavigation.tsx
```diff
@@ -24,7 +24,7 @@ export const MobileTopNavigation = () => {
   const authState = useAuth();
 
   const { navigation, options, getProfileMenuItems } = context;
-  const { isAuthenticated, profile, isAuthEnabled } = authState;
+  const { isAuthenticated, profile, isAuthEnabled, login } = authState;
```
mosch, line 27: isn't login an no-op now?

mosch:
Here's my [alternative approach.](https://github.com/zuplo/zudoku/pull/1677) Instead of complicating the interface, let's use it as intended. 

1. The `signUp`, `signIn`, and `logout` methods each perform specific actions and are solely responsible for managing the login process.
2. The Supabase Auth Plugin introduces its own routes and UI for login, rather than relying on the default implementation.

To make this work, I added a missing method to the interface: `navigate()`, which enables in-app navigation.

mosch:
Closed in fav of https://github.com/zuplo/zudoku/pull/1677

PR #1679, opened 2025-11-04
Add copy button when viewing API response in playground

ankitpandey2708:


packages/zudoku/src/lib/plugins/openapi/playground/result-panel/ResponseTab.tsx
```diff
@@ -273,24 +278,47 @@ export const ResponseTab = ({
...
+                }
+                code={
+                  (view === "raw"
+                    ? body
+                    : view === "types"
+                      ? types.data?.lines.join("\n")
+                      : beautifiedBody) ?? ""
+                }
+              />
+              <button
```
dan-lee: Can we use `Button` component with variant `ghost` here maybe?

dan-lee:
Sorry you have to rebase I had to make some changes there 😵‍💫

ankitpandey2708:
@dan-lee done.

dan-lee:
@ankitpandey2708 hmm looks like some things are missing now, no? I only see the imports and hook calls

ankitpandey2708:
@dan-lee its working fine after rebasing. u can merge.

mosch:
This PR is still not complete. The button does not get used.

PR #1732, opened 2025-11-21
feat: CLI `generate examples`

zeucapua:
Used inside a Zudoku project, generate examples for paths that don't have any in a given OpenAPI schema file. The CLI will have the following format:

```
zudoku generate examples --schema ./shipments.json --mode manual --paths /shipping /shipment/{id}
```

`generate` will be the top level command with `examples` being a subcommand. Written this way, there is room to generate more types of data if need be in the future. There are 3 flags:
- `--schema`: `-s` for short, **required**, specifies the relative path to an OpenAPI schema that will be modified in generation
- `--mode`: `-m` for short, optional, the method of creating the mock data with 3 options
  - `ai`: can pass in an OpenAI key to generate more realistic data, can be given an additional prompt for more accuracy (e.g. "only generate US based addresses")
  - `manual`: user will go through each endpoint and type in mock data per property for more control
  - `local`: similar to AI, but uses libraries like [faker](https://github.com/faker-js/faker) for more "generic" mock data
- `--paths`: `-p` for short, optional, used to only generate data for specific routes, useful when dev teams add a new endpoint within existing schemas

WIP

### DIFF
--- a/docs/pages/docs/configuration/authentication-firebase.md
+++ b/docs/pages/docs/configuration/authentication-firebase.md
@@ -50,7 +50,7 @@ export default {
     measurementId: "G-12W6TTNR75",
     // Optional: specify which providers to show in the sign-in UI
     // Available providers: "google", "github", "facebook", "twitter",
-    // "microsoft", "apple", "yahoo", "password", "phone", "emailLink"
+    // "microsoft", "apple", "yahoo", "password", "phone"
     // If not specified, all enabled providers in Firebase will be available
     providers: ["google", "github", "password"],
     // Optional: configure redirect URLs after authentication
--- a/docs/vite.config.ts
+++ b/docs/vite.config.ts
@@ -1,6 +1,9 @@
 import reactDocgenTypescript from "@joshwooding/vite-plugin-react-docgen-typescript";
 
 export default {
+  optimizeDeps: {
+    include: ["motion/react"],
+  },
   plugins: [
     reactDocgenTypescript({
       savePropValueAsString: true,
--- a/package.json
+++ b/package.json
@@ -39,8 +39,8 @@
       "use-sidecar": ">=1.1.3",
       "@babel/runtime": ">=7.26.10",
       "@babel/helpers": ">=7.26.10",
-      "react": "19.2.0",
-      "react-dom": "19.2.0"
+      "react": "19.2.1",
+      "react-dom": "19.2.1"
     },
     "patchedDependencies": {
       "decode-named-character-reference@1.0.2": "patches/decode-named-character-reference@1.0.2.patch",
@@ -62,5 +62,5 @@
       "check:other"
     ]
   },
-  "packageManager": "pnpm@10.20.0"
+  "packageManager": "pnpm@10.24.0"
 }
--- a/packages/create-zudoku/package.json
+++ b/packages/create-zudoku/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-zudoku",
-  "version": "0.65.3",
+  "version": "0.66.1",
   "keywords": [
     "react",
     "zudoku"
--- a/packages/zudoku/package.json
+++ b/packages/zudoku/package.json
@@ -1,6 +1,6 @@
 {
   "name": "zudoku",
-  "version": "0.65.3",
+  "version": "0.66.1",
   "type": "module",
   "homepage": "https://zudoku.dev",
   "repository": {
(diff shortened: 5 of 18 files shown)


examples/with-openapi-json/openapi.json
```diff
@@ -46,7 +46,14 @@
           "description": "Update an existent pet in the store",
           "content": {
             "application/json": {
-              "schema": { "$ref": "#/components/schemas/Pet" }
+              "schema": { "$ref": "#/components/schemas/Pet" },
+              "example": {
+                "name": "Jeremy",
+                "photoUrls": [
+                  "https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fjooinn.com%2Fimages%2Fdog-67.jpg&f=1&nofb=1&ipt=eefaf288694e06bdd2e57a0b4cc1e118583692685e6880e582f6ddfc13636c35"
```
mosch, line 53: 🐶

mosch:
Hey @zeucapua looking good! 

I looked at your findings and realized that the example can be defined in many places, so I think here is what we want:

- Allow to create **response examples** and **body exmaples**
 - If the response is defined on a referenced schema, we should ask if the example should go to the schema (default) or to the response

We want to replace this part where we generated examples on the fly based on the schema. We should not to that:

https://github.com/zuplo/zudoku/blob/main/packages/zudoku/src/lib/plugins/openapi/util/generateSchemaExample.ts

dan-lee:
Hey @zeucapua, thank you for this contribution 🤩 ! The `generate examples` CLI feature is a great
addition.

I know it's still a draft but some things that came to my mind:

**Use Existing Utilities Instead of Manual Traversal**

The implementation uses nested `Object.entries().map()` to manually traverse the OpenAPI schema.
Consider using the existing `traverse` utility:

```ts
import { traverse } from "../../lib/util/traverse.js";

const processedSchema = traverse(schema, (node, path) =>
  needsExample(node, path) ? addExample(node) : node,
);
```

**Handle References and File Format Support**

The implementation currently:

- Only handles JSON files (missing YAML support)
- Doesn't handle `$ref` references in schemas

How we could approach this:

For handling `$ref` references, we need to dereference to understand what the schema looks like (to generate appropriate examples), then write examples back to the original file.

I think we could use `$RefParser` (from `@apidevtools/json-schema-ref-parser`) for dereferencing, then parse the original file to write back.

For YAML files, the `yaml` library's `parseDocument` can preserve comments, indentation, property order, and other formatting (which is nice since authors often include comments for a purpose):

```typescript
import { $RefParser } from "@apidevtools/json-schema-ref-parser";
import { parseDocument } from "yaml";

// Dereference to understand the schema
const dereferenced = await new $RefParser().dereference(filePath);

// Collect ...

// Parse original file to preserve formatting
const doc = parseDocument(await fs.readFile(filePath, "utf-8"));

// Path comes from collected paths during traversal (e.g., ["components", "schemas", "User"])
doc.getIn(["components", "schemas", "User"]).set("example", generatedExample);
await fs.writeFile(filePath, doc.toString());
```

I think for JSON files, formatting with `JSON.stringify` should be fine as there are no comments or other formatting to preserve.

### User control over example placement:

As @mosch mentioned when a response uses a referenced schema, we could prompt the user whether to add examples to:

- The referenced schema in `components/schemas` (affects all uses)
- The specific response location (only affects this endpoint)

and/or we could have something like a `--strategy` flag:

```bash
# Default: add examples to referenced schemas in components
zudoku schema generate examples --input api.yaml --strategy reference

# Add examples inline at each usage location
zudoku schema generate examples --input api.yaml --strategy inline

# Ask user for each schema (interactive)
zudoku schema generate examples --input api.yaml --strategy ask
```

**Handle OpenAPI Version Differences**

OpenAPI 3.0 and 3.1 have different ways of defining examples:

- **OpenAPI 3.0:** Uses `example` (singular) field
- **OpenAPI 3.1:** Uses `examples` (plural) as an object with named examples

This needs to be addressed. You can either:

1. Support only OpenAPI 3.1+ and require users to upgrade first
2. Handle both versions explicitly by detecting the OpenAPI version:

```typescript
const isOpenAPI3_1 = schema.openapi?.startsWith("3.1");
const exampleField = isOpenAPI3_1 ? "examples" : "example";
```

**Command Structure**

The current structure is:

```bash
zudoku generate examples --schema ./openapi.json
```

What if we had a `schema` namespace to group all schema related operations:

```bash
zudoku schema generate examples --input ./openapi.json --mode manual
zudoku schema validate --input ./openapi.json
zudoku schema bundle --input ./openapi.json --output ./bundled.json
zudoku schema upgrade --input ./openapi.json
```

This would align with the vision of having a utility belt for dealing with OpenAPI schemas in the future. It groups related schema operations together and allows for future schema commands (`validate`, `bundle`, `upgrade`) that make explicit what `SchemaManager` does on-the-fly during the build.

**Some additional flags we could add**

- `--dry-run` - Preview what examples would be added without modifying files
- `--output <file>` - Write to a new file instead of modifying the original

---

Sorry for the lengthy feedback, but I believe this could be the foundation for our OpenAPI schema toolkit, so it's important we get it right from the start.

PR #1743, opened 2025-11-24
fix: Playground sets undefined body for "GET" and "HEAD"

jpmcb:
Ran into an issue in the playground where an error message would be displayed:

```
Request failed
Request constructor: HEAD or GET Request cannot have a body.
```

This seems to be because we pass along a `body` to the fetch `Request` even if there is no body. This patch checks for the method being `GET` or `HEAD` and sets the body to undefined if so.


jpmcb:
Confirmed fixed on deployed Cosmo cargo:

Screenshot 2025-11-25 at 7 56 05 AM

Normal Cosmo cargo fails with error: https://www.cosmocargo.dev/api-shipments/shipment-management?lang=shell

image

PR #1751, opened 2025-11-30
Add OpenAPI title to Pagefind results

### LINKED ISSUE #1600: Display current API in Pagefind results when multiple APIs indexed
Currently if you have multiple API specs indexed in your Zudoku instance, and you search in Pagefind, you'll get results from across multiple APIs but no way of knowing which API the result comes from. This is a particular problem for things like schemas, where you have multiple sets of schema results that just say "Schemas".

In certain cases, like the example below, where you may have similar or even identical endpoints across APIs for things like healthchecks, this makes it difficult to know which API clicking through will take you to.

Image

---

To help with this it would be great to include the title of the API in the subheader of the results, when Zudoku is set up to support multiple API specs.

This could look something like this:

<table>
  <tr><th>Single API</th><th>Multiple APIs</th></tr>
  <tr>
    <td valign="top"></td>
    <td valign="top"></td>
  </tr>
</table>

I think this should be pretty straightforward, by exposing the value of the `CategoryHeader` which contains the API title as an additional piece of PageFind metadata.

If this is something that seems reasonable I'm happy to look at contributing a PR.

dan-lee:
Adds the title of the OpenAPI document to Pagefind search results. I've named it `category` so other search results can benefit from this in the future

Fix #1600


packages/zudoku/src/lib/plugins/search-pagefind/ResultList.tsx
```diff
@@ -98,7 +103,7 @@ export const ResultList = ({
                     <div className="flex flex-col items-start gap-2 ms-2.5 ps-5 border-l border-muted-foreground/50">
                       <span className="font-bold">{subResult.title}</span>
                       <span
-                        className="text-[13px] [&_mark]:bg-primary [&_mark]:text-primary-foreground"
+                        className="line-clamp-2 text-[13px] [&_mark]:bg-primary [&_mark]:text-primary-foreground"
```
dan-lee, line 106: I think the search results can become too crowded so I've added a line clamp here.

PR #1752, opened 2025-11-30
Fix internal link takes over sidebar section

### LINKED ISSUE #1556: Link between sections
It would be helpful to be able to create "external" links to a different section of the site. For example, i have two top level categories and I want to put a document on category A that links to the document in category B. Currently if you create a item that is type `link` that links to a document on the same site the item will open with the same sidebar. I think the behaviour should be that if you create a `link` to a document/page/etc. within the same site it should do a "full" navigate so that the page opens to the category where the document (or custom page, etc) is set.

Minimal configuration below:

```ts
import type { ZudokuConfig } from "zudoku";

const config: ZudokuConfig = {
  navigation: [
    {
      type: "category",
      label: "Category A",
      items: [
        "foo", // This is a `type: doc`
      ],
    },
    {
      type: "category",
      label: "Category B",
      items: [
        {
          type: "link",
          label: "Foo",
          to: "/foo",
        },
      ],
    },
  ],
};

export default config;

```

I am not sure if this can work by default (it might break other uses of the `link` type). If not maybe this is a setting on the object `external: true` or something.

I think it might be good to put the external icon  Image on these type of links as well to show that the content is in another section.

dan-lee:
Fix #1556

### DIFF
--- a/packages/zudoku/src/lib/components/context/ZudokuContext.ts
+++ b/packages/zudoku/src/lib/components/context/ZudokuContext.ts
@@ -45,8 +45,6 @@ const getItemPath = (item: NavigationItem) => {
       return joinUrl(item.path);
     case "category":
       return item.link ? joinUrl(item.link.path) : undefined;
-    case "link":
-      return item.to;
     case "custom-page":
       return joinUrl(item.path);
     default:


dan-lee:
Superseded by #1862

PR #1787, opened 2025-12-05
Schema download url for latest version

rikard-swahn:
Currently, there is no URL for getting the latest version, only URL:s with specific version numbers. This PR makes a /latest/ version available as a download URL, in addition to the versioned URLs. The latest version is defined as the schema first in the apis.*.input array

Also, current download URL for schemas changes whenever the schema file name changes. This PR replaces filename with "schema.<extension>".</extension>

This allows for static URLs for services, so that integration with the docs becomes more stable.

Testing:
- In the "Download schema" button, choose "Open in new tab". 
Expected: URL has "schema.<extension>" instead of schema file name, e.g.  `/api-shipments/1.0.0/schema.json`.

- Replace version number with `latest` in URL, e.g. `/api-shipments/latest/schema.json`.
Expected: The latest version of the schema is returned.

### DIFF
--- a/packages/zudoku/src/lib/plugins/openapi/OperationList.tsx
+++ b/packages/zudoku/src/lib/plugins/openapi/OperationList.tsx
@@ -256,10 +256,7 @@ export const OperationList = ({
     typeof input === "string"
       ? type === "url"
         ? input
-        : (() => {
-            const extension = getFileExtension(input);
-            return joinUrl(path, version, `schema${extension}`);
-          })()
+        : joinUrl(path, version, `schema${getFileExtension(input)}`)
       : undefined;
 
   return (


dan-lee:
Thanks @rikard-swahn for your contribution! In general I'm fine with this, however one concern I have:

The current implementation physically duplicates schema files in the build output. For example the cosmo-cargo project I see this in `dist` folder:

```
api-shipments/1.0.0/schema.json  (117KB)
api-shipments/latest/schema.json (117KB, identical copy)
```

For projects with large schemas or many APIs this could noticeably increase the total build size.
Maybe we could symlink to the `latest` schema?

rikard-swahn:
> Thanks @rikard-swahn for your contribution! In general I'm fine with this, however one concern I have:
> 
> The current implementation physically duplicates schema files in the build output. For example the cosmo-cargo project I see this in `dist` folder:
> 
> ```
> api-shipments/1.0.0/schema.json  (117KB)
> api-shipments/latest/schema.json (117KB, identical copy)
> ```
> 
> For projects with large schemas or many APIs this could noticeably increase the total build size. Maybe we could symlink to the `latest` schema?

Symlinks are now created for /latest which point to latest version.

rikard-swahn:
I set the PR to draft. I think the semver comparison should be removed, and just define the latest version as the first one in the array of schemas.

packages/zudoku/src/lib/plugins/openapi/DownloadSchemaButton.tsx
```diff
@@ -86,6 +88,18 @@ export const DownloadSchemaButton = ({
...
+              const absoluteUrl = new URL(
+                downloadUrlLatest,
+                window.location.href,
+              ).href;
+              copyToClipboard(absoluteUrl);
+            }}
+          >
+            <CopyIcon size={14} />
+            Copy latest URL
+          </DropdownMenuItem>
```
dan-lee: I don't think we should be cross-referencing to other schemas, could just switch to the latest version. Feels a bit out of place

packages/zudoku/src/lib/plugins/openapi/OperationList.tsx
```diff
@@ -151,6 +151,27 @@ const OperationsForTagQuery = graphql(/* GraphQL */ `
...
+  path: string | undefined,
+  version: string | undefined,
+) {
+  return typeof input === "string"
+    ? type === "url"
+      ? input
+      : (() => {
+          const extension = getFileExtension(input);
+          return joinUrl(path, version, `schema${extension}`);
+        })()
```
dan-lee: I think this IIFE is not very clean. Why not inline it like


```suggestion
      : joinUrl(path, version, `schema${getFileExtension(input)}`)
```

packages/zudoku/src/lib/plugins/openapi/OperationList.tsx
```diff
@@ -246,13 +267,8 @@ export const OperationList = ({
...
   const helmetTitle = [tagTitle, title].filter(Boolean).join(" - ");
-
-  const downloadUrl =
-    typeof input === "string"
-      ? type === "url"
-        ? input
-        : joinUrl(path, version, input.split("/").pop())
-      : undefined;
+  const downloadUrl = getDownloadUrl(input, type, path, version);
+  const downloadUrlLatest = getDownloadUrl(input, type, path, "latest");
```
dan-lee: Again, as above I don't think we should have a link to latest in here

packages/zudoku/src/lib/plugins/openapi/OperationList.tsx
```diff
@@ -151,6 +151,27 @@ const OperationsForTagQuery = graphql(/* GraphQL */ `
 
 const LAZY_OPERATION_LIST_THRESHOLD = 30;
 
+const getFileExtension = (filename: string): string => {
+  const lastDotIndex = filename.lastIndexOf(".");
+  return lastDotIndex !== -1 ? filename.slice(lastDotIndex) : "";
+};
```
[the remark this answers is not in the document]
dan-lee: `path.extname` is not valid here because it's not Node.js context

packages/zudoku/src/lib/plugins/openapi/DownloadSchemaButton.tsx
rikard-swahn: Yes, but if you are on the latest version and then select "Open in new tab", an url with the latest version in it is used (as it should). If I want an URL which always points to the latest version, what would I do? How would the user know there is such an URL?

I thought about adding a "Latest (1.0.0)" version in the version selector (in addition to "1.0.0"), but that becomes even more confusing I think.

dan-lee: Good question, I am afraid I can't really give an answer. You mentioned that it's helpful for services to consume, maybe it's just ok that this is a URL that has to be known instead of showing it everywhere?

dan-lee:
Turns out symlinks are more fragile than I thought across different environments 😞 

1. Windows: Symlink creation requires admin privileges or Developer Mode, which would probably break builds for quite some users
1. Static hosting: Services like S3, GitHub Pages, and Vercel don't handle symlinks very well, leading to 404s on `/latest/` URLs

Maybe we should revert to your initial approach since it's much more robust. Sorry for the back and forth, but I'd like to get this right :)

packages/zudoku/src/lib/plugins/openapi/DownloadSchemaButton.tsx
rikard-swahn: Ok, removed "latest" from UI

rikard-swahn:
> Turns out symlinks are more fragile than I thought across different environments 😞
> 
> 1. Windows: Symlink creation requires admin privileges or Developer Mode, which would probably break builds for quite some users
> 2. Static hosting: Services like S3, GitHub Pages, and Vercel don't handle symlinks very well, leading to 404s on `/latest/` URLs
> 
> Maybe we should revert to your initial approach since it's much more robust. Sorry for the back and forth, but I'd like to get this right :)

Reverted

PR #1811, opened 2025-12-13
Change spread order so authentication-specific basePath can override the global value:

ankitpandey2708:
This fix ensures:

    When authentication.basePath is specified, it overrides the global basePath (correct behavior)
    When authentication.basePath is not specified, the global basePath is used (backward compatible)
    When neither is specified, basePath remains undefined (correct default behavior)


dan-lee:
There's no authentication-specific `basePath`, this is a global configuration.

PR #1812, opened 2025-12-13
fix: renderBuiltUrl omits basePath when generating CDN asset URLs

ankitpandey2708:
The joinUrl utility function already handles undefined or empty string values correctly, so this fix will work whether or not basePath is configured.


dan-lee:
Hey @ankitpandey2708 Is this really an issue? It seems to work for us without problems. The file path should include the base path already because it's put into that folder anyway.

PR #1813, opened 2025-12-13
Fix: Implement React Router path composition in routesToPaths for nested routes

ankitpandey2708:


dan-lee:
Hey @ankitpandey2708, thanks for the PR. Can you explain what this PR intends to fix?

ankitpandey2708:
@dan-lee 

## Background
The current implementation doesn't track parent paths when recursively processing route children, which violates React Router's standard path composition behavior. While this doesn't currently break prerendering (because all plugins use `joinUrl()` to pre-build absolute paths before creating routes), it creates a footgun for plugin developers and users who might use the standard React Router pattern of relative child paths.

packages/zudoku/src/vite/prerender/utils.ts
```diff
@@ -15,13 +15,18 @@ export const routesToPaths = (routes: RouteObject[]) => {
       }
 
       if (route.path) {
-        paths.push(route.path.startsWith("/") ? route.path : `/${route.path}`);
-      }
-      if (route.children) {
-        addPaths(route.children);
+        const fullPath = route.path.startsWith("/")
+          ? route.path
+          : parentPath + "/" + route.path;
```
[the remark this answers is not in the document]
ankitpandey2708: done

packages/zudoku/src/vite/prerender/utils.ts
```diff
@@ -1,8 +1,28 @@
...
+
+  // If child is empty, return parent as-is
+  if (!normalizedChild) {
+    return normalizedParent;
+  }
+
+  // Join with single slash and ensure leading slash
+  const joined = `${normalizedParent}/${normalizedChild}`;
+  return joined.startsWith("/") ? joined : `/${joined}`;
+};
```
dan-lee: Let's reuse the already existing [`joinUrl`](https://github.com/zuplo/zudoku/blob/main/packages/zudoku/src/lib/util/joinUrl.ts) here.

packages/zudoku/src/vite/prerender/utils.ts
```diff
@@ -15,10 +35,21 @@ export const routesToPaths = (routes: RouteObject[]) => {
...
+        const fullPath = route.path.startsWith("/")
+          ? route.path
+          : joinPaths(parentPath, route.path);
+        // Normalize the path before pushing to avoid multiple consecutive slashes
+        const normalizedPath = fullPath.replace(/\/+/g, "/");
+        paths.push(
+          normalizedPath.startsWith("/")
+            ? normalizedPath
+            : `/${normalizedPath}`,
+        );
```
dan-lee: I think we could simplify it to:

```tsx
paths.push(
  route.path.startsWith("/") ? route.path : joinUrl(parentPath, route.path),
);
```

PR #1814, opened 2025-12-13
Use safe string serialization when embedding the URL in JavaScript

ankitpandey2708:


dan-lee:
Thanks for the PR! In this case the URL comes from the app's own route definitions at build time, not from user input, so there's no realistic attack vector here.

PR #1815, opened 2025-12-13
fix : API key deletion requests fail with authentication errors

ankitpandey2708:
The fix matches the pattern used in rollKey and updateConsumer functions in the same file.

### DIFF
--- a/packages/zudoku/src/lib/plugins/api-keys/index.tsx
+++ b/packages/zudoku/src/lib/plugins/api-keys/index.tsx
@@ -92,9 +92,7 @@ const createDefaultHandler = (
           method: "DELETE",
         },
       );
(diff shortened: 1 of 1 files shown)


packages/zudoku/src/lib/plugins/api-keys/index.tsx
```diff
@@ -92,8 +92,9 @@ const createDefaultHandler = (
           method: "DELETE",
         },
       );
-      await context.signRequest(request);
-      const response = await fetch(request);
+      const response = await fetch(
+        await context.signRequest(request),
+      );
```
dan-lee:
```suggestion
      const response = await fetch(await context.signRequest(request));
```

PR #1817, opened 2025-12-13
Fix OAuth callback handler reusing cached redirect URLs from previous flows

ankitpandey2708:


ankitpandey2708:
## Problem

The `CallbackHandler` component uses a static React Query cache key (`["oauth-callback"]`) combined with a global 5-minute `staleTime`. This causes multiple OAuth authentication flows within 5 minutes to incorrectly reuse the redirect URL from the first flow instead of using the intended destination for each subsequent flow.

**Impact:**
- Users completing a second OAuth flow within 5 minutes are redirected to the wrong page
- Affects scenarios like: signing in from different tabs, clicking back and signing in again, or testing OAuth flows during development

**Example:**
1. User completes OAuth at 12:00, intending to go to `/docs/api-reference` ✅
2. User completes OAuth at 12:02, intending to go to `/settings` 
3. User is incorrectly redirected to `/docs/api-reference` ❌

ankitpandey2708:
@dan-lee

dan-lee:
Thanks!

PR #1818, opened 2025-12-13
Fix pagefind meta tag not rendering for protected routes during SSR

ankitpandey2708:
The `isProtectedRoute` flag was tied to `shouldBypass`, so it was always `false` during SSR/prerendering. This prevented the pagefind meta tag from rendering for protected routes.
                                                                                                                                     
Refactored `RouteGuard` to:

- Compute `isProtectedRoute` independently of bypass flag
- Early return in bypass mode to skip auth and render meta tag
- Keep auth validation only for normal (non-bypass) flow
                                                                                                                                     
This separates SSR and normal modes cleanly instead of mixing `shouldBypass` into multiple conditions.

_edited by @dan-lee to keep it up-to-date_

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,9 @@ updates:
   # Maintain dependencies for npm
   - package-ecosystem: "npm"
     directory: "/"
+    cooldown:
+      # See https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#cooldown-
+      default-days: 3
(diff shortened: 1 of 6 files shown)


dan-lee:
Protected content was being indexed during prerendering but wasn’t tagged as “protected” in the search index because bypassing auth also incorrectly cleared the `isProtectedRoute` flag.

dan-lee:
Thanks for your contribution @ankitpandey2708!

PR #1819, opened 2025-12-14
Fix : duplicate auth state initialization in Clerk provider

ankitpandey2708:


dan-lee:
Hey @ankitpandey2708 I am not sure I understand this PR

ankitpandey2708:
@dan-lee 

## Problem

The Clerk provider initializes auth state in two places:

1. **IIFE (lines 32-57)**: Runs immediately when plugin is created, checks `clerkApi.user` and sets auth state
2. **`initialize()` method**: Runs when framework is ready, checks `clerk.session` and sets auth state again

This creates race conditions:
- State updates happen before framework is ready
- Different checks (`user` vs `session`) may produce inconsistent results  
- IIFE never sets `isAuthenticated: false` when logged out
- Duplicate state updates with slightly different data

## Solution

Remove lines 32-57 so the IIFE only loads the Clerk SDK. All auth state initialization happens in `initialize()` at the proper time.

**Before:** SDK load → set state → framework ready → set state again  
**After:** SDK load → framework ready → set state once

This ensures a single source of truth with proper handling of both authenticated and unauthenticated states.

mosch:
@ankitpandey2708 where would the "other" set state happen? I can only see the one you are removing in this PR.

ankitpandey2708:
@mosch There are two setLoggedIn calls: one in ensureLoaded (which I removed) and one in initialize (which remains)

mosch:
Got you - thanks for clarification 😵‍💫

PR #1821, opened 2025-12-14
fix : Bypass render status ignored for protected routes in prerender worker

ankitpandey2708:


packages/zudoku/src/vite/prerender/worker.ts
```diff
@@ -77,6 +83,7 @@ const renderPage = async ({ urlPath }: WorkerData): Promise<WorkerResult> => {
     html = fileResponse.buffer;
   }
 
+  // Check fileResponse status for both protected and non-protected routes
```
dan-lee:
```suggestion
```

dan-lee (approved):

PR #1822, opened 2025-12-14
fix : Prevent path traversal in MCP file processing

ankitpandey2708:


dan-lee:
Thanks for thinking about security here! However, this runs at build time on developer-authored config files, so if someone could inject malicious paths into the OpenAPI spec, they'd already have full project access anyway.

Issue #1831, opened 2025-12-17
Formatting issue with Tables


skeyby:
There's a formatting issue with Tables in markdown pages.

The documentation says that Zudoku supports Github Flavoured Markdown but there's an inconsistency with new lines inside table cells.

This basic example summarise it:

Sample table:

```
| Heading 1 | Heading 2 |
|---|---|
|This line <br>should be cut|This one not|
```

Github correct render:

| Heading 1 | Heading 2 |
|---|---|
|This line <br>should be cut|This one not|

Zudoku wrong render

<img width="800" height="103" alt="Image" src="https://github.com/user-attachments/assets/6d488748-f94d-4983-af0a-37b1dc90df66" />

ankitpandey2708:
<img width="1919" height="515" alt="Image" src="https://github.com/user-attachments/assets/2912662a-37fd-48b1-bc45-5f01d5b39762" />

@skeyby have created a PR

skeyby:
Do you think the PR will land soon in one of the upcoming versions?

ankitpandey2708:
@dan-lee pls push

dan-lee:
@skeyby Can you use `.mdx` for now? We have to make sure first that this won't break in the ecosystem

skeyby:
hi @dan-lee - yes, if the page is renamed to .mdx it works as intended.

Can I expect any "impact" (?) of any kind by using an .mdx page for basic .md stuff?

I mean, am I enabling something else without being aware of it?

dan-lee:
@skeyby MDX is a superset of markdown so your content works the same. Main additions: import React components, use JSX inline, export metadata.
One caveat is that MDX is stricter with syntax. Angle brackets and curly braces may need escaping since they're interpreted as JSX. If you hit weird parsing errors, that's usually why.

skeyby:
Cool. Let me know if and when you'll address it on the .md side then.

mosch:
Please use `mdx` instead of `md` going forward if you want to have support for HTML Tags.

PR #1832, opened 2025-12-17
feat: add HTML support in markdown tables via rehype-raw

ankitpandey2708:
Added rehype-raw plugin to MDX plugin configuration with passThrough options to support HTML elements like <br /> within markdown tables while maintaining MDX compatibility.

The rehype-raw configuration passes through MDX-specific nodes (mdxFlowExpression, mdxJsxFlowElement, mdxTextExpression, mdxjsEsm, etc.) to prevent compilation errors while enabling HTML in markdown content.

This allows users to use HTML elements within markdown tables:
- Line breaks with <br />
- Inline styles with <span>
- Other HTML elements mixed with markdown

MDX uses JSX syntax, which requires self-closing tags to use the /> format:

    ✅ Use <br />, <hr />, 
    ❌ Don't use <br>, <hr>, 

## Summary by CodeRabbit

* **Chores**
  * Improved MDX content processing to preserve and pass through raw HTML within MDX constructs, ensuring embedded HTML and mixed markdown/JSX fragments render correctly across pages and components.

### DIFF
--- a/packages/zudoku/src/vite/plugin-mdx.ts
+++ b/packages/zudoku/src/vite/plugin-mdx.ts
@@ -5,6 +5,7 @@ import withTocExport from "@stefanprobst/rehype-extract-toc/mdx";
 import type { Root as HastRoot } from "hast";
 import { toString as hastToString } from "hast-util-to-string";
 import rehypeMdxImportMedia from "rehype-mdx-import-media";
+import rehypeRaw from "rehype-raw";
 import rehypeSlug from "rehype-slug";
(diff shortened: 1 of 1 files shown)


ankitpandey2708:
@dan-lee

ankitpandey2708:
closes https://github.com/zuplo/zudoku/issues/1831

dan-lee:
We need to test this thoroughly before we can merge it. Isn't this potentially breaking? Why can't one just use MDX?

PR #1834, opened 2025-12-17
fix : Improve React version resolution for monorepo setups

ankitpandey2708:
### Problem
When developing in the Zudoku monorepo, running `pnpm dev` in example directories fails with "module not found" errors.

**Steps to reproduce:**
```bash
# Root directory
pnpm install && pnpm --filter zudoku build

# Any example directory
cd examples/with-config
pnpm dev  # ❌ Fails
```

### Root Cause
The previous implementation used hardcoded path construction to locate React packages:
```bash
const nodeModulesPath = fileURLToPath(new URL("../../../..", import.meta.url));
const reactPath = path.join(nodeModulesPath, "react/package.json");
```

### DIFF
--- a/packages/zudoku/src/cli/common/version-check.ts
+++ b/packages/zudoku/src/cli/common/version-check.ts
@@ -28,11 +28,20 @@ export const warnPackageVersionMismatch = async () => {
   const packageJson = getPackageJson(
     fileURLToPath(new URL("../../../package.json", import.meta.url)),
   );
-  const nodeModulesPath = fileURLToPath(
-    new URL("../../../..", import.meta.url),
-  );
-  const reactPath = path.join(nodeModulesPath, "react/package.json");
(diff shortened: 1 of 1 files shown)


dan-lee:
I can't reproduce this -- `pnpm dev` works in example directories for me.
Note the recommended way to run examples is `nx run with-config:dev` (with the env var `ZUDOKU_INTERNAL_DEV=1` set). See more in https://github.com/zuplo/zudoku/blob/main/CONTRIBUTING.md

Also, what Copilot said :)

> Using `require.resolve` in an ES module context (package.json has "type": "module") is not recommended and may cause issues in some Node.js versions. Consider using `import.meta.resolve` instead, which is the ES module equivalent. Other parts of the codebase already use this pattern (e.g., in vite/config.ts).

PR #1882, opened 2026-01-19
Add version config support for file-based API schemas

dan-lee:
Allow file type APIs to use VersionConfig with path/label overrides, matching URL type behavior.

Surfaced in https://discord.com/channels/848913990360629268/1461301705986543800

### DIFF
--- a/packages/zudoku/src/lib/plugins/openapi/interfaces.ts
+++ b/packages/zudoku/src/lib/plugins/openapi/interfaces.ts
@@ -7,8 +7,8 @@ type DynamicInput = () => Promise<unknown>;
 
 export type VersionedInput<T> = Array<{
   path: string;
-  version: string;
-  downloadUrl: string;
+  version?: string;
+  downloadUrl?: string;
   label?: string;
   input: T;
 }>;
--- a/examples/cosmo-cargo/zudoku.config.tsx
+++ b/examples/cosmo-cargo/zudoku.config.tsx
@@ -238,7 +238,11 @@ const config: ZudokuConfig = {
     {
       type: "file",
       input: [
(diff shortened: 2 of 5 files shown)


packages/zudoku/src/lib/plugins/openapi/interfaces.ts
```diff
@@ -85,5 +87,8 @@ export type OasPluginConfig = BaseOasConfig & OasSource;
 export type OasPluginContext = BaseOasConfig &
   ContextOasSource & {
     version?: string;
-    versions: Record<string, { path: string; label: string }>;
+    versions: Record<
+      string,
+      { path: string; label: string; downloadUrl?: string }
```
egrimstad, line 92: I don't think the `downloadUrl` is ever filled now? In my changes it is filled out in the `OasProvider`: https://github.com/zuplo/zudoku/pull/1883/changes#diff-2776e8253437ba15df70961dde051b438d61ac04e8d4e7286b6b31527f1ede17R35

dan-lee: Good spot, forgot to push that

packages/zudoku/src/lib/plugins/openapi/interfaces.ts
```diff
@@ -7,6 +7,8 @@ type DynamicInput = () => Promise<unknown>;
 
 export type VersionedInput<T> = Array<{
   path: string;
+  version: string;
+  downloadUrl: string;
```
egrimstad: `OperationList.tsx` can be updated to use this, instead of creating the url itself.

egrimstad (approved):

PR #1892, opened 2026-01-23
Expose all data from OpenID `userInfo` 

Keshav-Motorq:
With the Auth0 provider, additional fields such as app_metadata or user_metadata returned from the /userinfo endpoint are not being mapped to the profile object, despite being included in the userinfo response.


mosch:
Thanks for this PR. We're happy to get this in, however the data should be put in `providerData`. The `providerData` is an open property that any provider specific data can be stored.

Keshav-Motorq:
> Thanks for this PR. We're happy to get this in, however the data should be put in `providerData`. The `providerData` is an open property that any provider specific data can be stored.

Thanks, will make the required change

mosch:
I checked the interface again and noticted we allow for any value on the profile object, so we're good to go.

mosch (approved):

PR #1902, opened 2026-01-27
feat: Add audioplayer in playground

picsoung:
## Summary

Adds inline audio playback for binary audio responses in the API playground. When an API returns audio content (e.g., `audio/mpeg`, `audio/wav`), users can now listen directly in the playground instead of downloading the file.

## Motivation

APIs returning audio (text-to-speech, audio generation, etc.) currently show a generic "Binary Content" message requiring users to download and open files externally. This breaks the testing flow.

This change lets developers immediately listen to audio responses in the playground, improving the experience for testing audio-based APIs.

## Changes

**Core Feature** (`packages/zudoku/src/lib/plugins/openapi/playground/`)

| File | Change |
|------|--------|
| `fileUtils.ts` | Added `isAudioContentType()` helper |
| `result-panel/AudioPlayer.tsx` | **New** - Audio player component |
| `result-panel/ResponseTab.tsx` | Import and render AudioPlayer for audio responses |

**Demo** (optional - can be excluded from PR)

| File | Change |
|------|--------|
| `vite/dev-server.ts` | Demo endpoint `/captain-speaking/welcome` |
| `examples/cosmo-cargo/` | Example API spec + audio file |
| `examples/cosmo-cargo/public/welcome-aboard.mp3 ` | Example of audio file |

# How it looks
Screenshot 2026-01-27 at 1 50 02 PM

## Technical Details

- Uses `URL.createObjectURL()` with proper cleanup in `useEffect` to prevent memory leaks
- Supports all `audio/*` MIME types
- Falls back to existing binary download UI for non-audio content

## Test Plan

1. Create an API endpoint returning audio (`Content-Type: audio/mpeg` or similar)
2. Make a request in the playground
3. Verify audio player appears with playback controls
4. Verify download button works
5. Verify non-audio binary responses still show download UI


dan-lee:
Nice feature! Could we use a CDN-hosted audio file instead of bundling the MP3? That way we don't need the dev-server endpoint either.

 OpenAPI supports operation-level servers, so this should do ([audio file](https://archive.org/download/testmp3testfile/mpthreetest.mp3)):

```json
{
  "/mpthreetest.mp3": {
    "get": {
      "servers": [
        {
          "url": "https://archive.org/download/testmp3testfile",
          "description": "Audio sample"
        }
      ],
      "tags": ["AI Operations"],
      "summary": "Sample audio playback",
      "description": "Demo endpoint for audio player feature.",
      "operationId": "getSampleAudio",
      "responses": {
        "200": {
          "description": "Audio file",
          "content": {
            "audio/mpeg": {
              "schema": { "type": "string", "format": "binary" }
            }
          }
        }
      }
    }
  }
}
```

wdyt?

picsoung:
Indeed @dan-lee, I didn't know that was a thing.

I went ahead and removed the mp3 file + the endpoint in the dev-server
and added the operation-level server pointing to that archive.org mp3 file.

PR #1947, opened 2026-02-10
Better sorting of categories in OpenAPI

robinmoneybird:
When `x-tagGroups` is defined in an OpenAPI schema, grouped tags always appear before ungrouped tags. This makes it impossible to produce an alphabetically sorted sidebar that mixes groups with standalone tags. This change alphabetically interleaves while preserving group nesting.

Changes:
- Change tag and tagGroup sorting to interleave alphabetically
- Extract tag-building logic into a testable buildTagCategories utility
- Propagate the option through Vite code generation for both file and URL API types

Testing:
```sh
cd packages/zudoku && pnpm test -- \
  src/lib/plugins/openapi/util/buildTagCategories.test.ts
```

Success: all 6 tests pass — covering no tag groups, alphabetical sorting, nested structure preserved, empty groups filtered, expandAllTags toggle, and empty input.

Failure: tests fail if alphabetical sorting breaks when interleaving is enabled, or if group wrapper structure is lost.


mosch:
Hey, thanks for this PR! I wonder if this should be the default instead of configuration?

robinmoneybird:
> I wonder if this should be the default instead of configuration?

I'll leave that decision up to you. I've chosen to default it to `false` so this change isn't breaking to other workflows. Currently, we're only using this as a stopgap until support is added for [the changes to enhanced tags](https://learn.openapis.org/specification/tags.html#enhanced-tags-in-openapi-32) in [OpenAPI version 3.2.0](https://spec.openapis.org/oas/latest.html#tag-object), which add proper hierarchy to tags.

mosch:
🎉 We’re loving this change! Always sorting the groups sounds like a fantastic default to us. Let’s skip the options and embrace the order! 

If someone wants differnt sorting, they can use the new `navigationRules` feature to change sorting. https://github.com/zuplo/zudoku/pull/1949

@robinmoneybird thanks again, would you be up for making the change to remove the option?

robinmoneybird:
> would you be up for making the change to remove the option?

Yes, no problem. The change has been made, let me know if you need something changed in review 🙂

mosch (approved):

PR #1959, opened 2026-02-12
Merge tags and tagGroups with the same name in sidebar

robinmoneybird:
Mimic the more permissive "Enhanced tags" behaviour from OpenAPI 3.2.0, by allowing a tag to have both direct operations and nested child tags from tagGroups. When a tag and x-tagGroup share the same name, they are now combined into a single sidebar category instead of appearing twice.

However, this current implementation does not use `x-displayName` as, these could still be different tags. Only the actual `name` of a tag is considered when checking name equality.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit tests for tag categorization and grouping logic covering various merging scenarios.

* **Improvements**
  * Enhanced tag and category merging strategy for improved organization of OpenAPI documentation categories.

### DIFF
--- a/packages/zudoku/src/lib/plugins/openapi/util/buildTagCategories.test.ts
+++ b/packages/zudoku/src/lib/plugins/openapi/util/buildTagCategories.test.ts
@@ -104,29 +104,11 @@ describe("buildTagCategories", () => {
     expect(contacts?.type).toBe("category");
 
     if (contacts?.type === "category") {
-      // Operations from the "Contacts" tag come first, then "Notes" as a child
+      // Operations from the "Contacts" tag come first, then "Notes" as a child.
+      // "Contacts" must not appear as a nested child of itself.
       expect(contacts.items).toHaveLength(2);
       expect(contacts.items[0]?.label).toBe("Contacts op");
       expect(contacts.items[1]?.label).toBe("Notes");
-    }
-  });
-
-  it("excludes self-reference when tagGroup includes its own name in tags", () => {
-    const tagCategories = new Map<string, NavigationItem>([
-      ["Contacts", makeTag("Contacts")],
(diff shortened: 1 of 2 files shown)


robinmoneybird:
When an OpenAPI spec has a tag and an x-tagGroups entry that share the same name (e.g. "Contacts"), they previously rendered as two separate sidebar entries — one with the tag's operations, and one wrapping the group's child tags. This merges them into a single category:  the tag's own operations appear first, followed by nested child tags from the group.        
                                                                                                                                                                                            
This is useful for specs that want to organize operations hierarchically — similar to how Enhanced Tags in OpenAPI 3.2.0 allow tags to have both direct paths and nested sub-tags. With this change, you can achieve the same structure today using x-tagGroups without ending up with duplicate sidebar entries.

### Example

```yml
  tags:
    - name: Contacts
    - name: Notes

  x-tagGroups:
    - name: Contacts
      tags:
        - Notes

  paths:
    /contacts:
      get:
        tags: [Contacts]
    /contacts/{id}/notes:
      get:
        tags: [Notes]
```

Before — two sidebar entries:

```text
▶ Contacts       ← from x-tagGroups, wrapping children
  ▶ Notes        ← tag nested within x-tagGroups
    GET /contacts/{id}/notes
▶ Contacts       ← direct tag outside x-tagGroups
  GET /contacts
```

After — single merged entry:

```text
▶ Contacts           ← merged tag + tagGroup
  GET /contacts      ← operations from the "Contacts" tag
  ▶ Notes            ← nested child tag from the group
    GET /contacts/{id}/notes
```

dan-lee:
Thanks for the contribution @robinmoneybird! 🚀
I rebased it and made some minor changes

dan-lee (approved):

PR #1962, opened 2026-02-12
docs: add client-side only caveat to protectedRoutes

dan-lee:
## Summary
- Adds caution callout to `protectedRoutes` docs clarifying that protection is client-side only since Zudoku is SSG
- Notes that all content is bundled into JS and visible regardless of auth state
- Mentions SSR with proper access control is in development

### DIFF
--- a/docs/pages/docs/configuration/protected-routes.md
+++ b/docs/pages/docs/configuration/protected-routes.md
@@ -11,6 +11,18 @@ your [Zudoku configuration](./overview.md). This requires [authentication](./aut
 be configured. The property supports two formats: a simple array of path patterns, or an advanced
 object format with custom authorization logic.
 
+:::caution[Client-side protection only]
+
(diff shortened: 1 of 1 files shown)


docs/pages/docs/configuration/protected-routes.md
```diff
@@ -11,6 +11,18 @@ your [Zudoku configuration](./overview.md). This requires [authentication](./aut
 be configured. The property supports two formats: a simple array of path patterns, or an advanced
 object format with custom authorization logic.
 
+:::caution[Client-side protection only]
```
joshtwist, line 14: I would probably make this a little bit less shouty and not in a big caution box and just put it in a note. Something like this: 

*Note*: protectedRoutes are client-side only. Do not rely on protectedRoutes to hide sensitive information. 

We are working on an additional offering that secures this data server-side.

PR #1968, opened 2026-02-12
Revert OpenAPI default tag sorting from #1947

dan-lee:
Reverts the default alphabetical sorting of tag categories introduced in #1947. Sorting by default changes the sidebar order for all APIs which is a breaking change.

The same behavior can be achieved using the upcoming `navigationRules` feature #1949.

### DIFF
--- a/packages/zudoku/src/lib/plugins/openapi/util/buildTagCategories.test.ts
+++ b/packages/zudoku/src/lib/plugins/openapi/util/buildTagCategories.test.ts
@@ -13,44 +13,6 @@ const makeTag = (label: string): NavigationItem => ({
 });
 
 describe("buildTagCategories", () => {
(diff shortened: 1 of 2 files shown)


dan-lee:
Hey @robinmoneybird thanks for your contribution in #1947; sorry for the back and forth on this. We suggested making it the default but it changes the sidebar order for all APIs which is more breaking than we anticipated.
So we ultimately have to revert your main change.

However you'll be able to get the same result in the next version using `navigationRules` (#1949):

```ts
navigationRules: [
  {
    type: "sort",
    match: "Your API", // match by label
    by: (a, b) => a.label.localeCompare(b.label),
  },
],
```

See [the preview docs for `navigationRules`](https://zudoku-dev-git-feat-navigation-rules.zuplosite.com/docs/guides/navigation-rules)

mosch (approved):

PR #1993, opened 2026-02-18
fix(env): Add `ZUPLO_PUBLIC_DISABLE_INJECT_ENDPOINT` to stop overwriting server URL

mosch:
## Summary by CodeRabbit

* **Chores**
  * Server configuration now prioritizes a new environment variable while remaining backward compatible with the previous setting.
* **Bug Fixes**
  * Public API documentation generation no longer injects a server endpoint when injection has been explicitly disabled, preventing unintended URL exposure.

### DIFF
--- a/packages/zudoku/src/app/env.ts
+++ b/packages/zudoku/src/app/env.ts
@@ -36,7 +36,7 @@ export const ZuploEnv = {
   },
 
   get serverUrl(): string | undefined {
-    return process.env.ZUDOKU_SERVER_URL ?? process.env.ZUPLO_SERVER_URL;
+    return process.env.ZUPLO_SERVER_URL;
   },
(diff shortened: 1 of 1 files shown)


packages/zudoku/src/app/env.ts
```diff
@@ -36,7 +36,7 @@ export const ZuploEnv = {
   },
 
   get serverUrl(): string | undefined {
-    return process.env.ZUPLO_SERVER_URL;
+    return process.env.ZUDOKU_SERVER_URL ?? process.env.ZUPLO_SERVER_URL;
```
dan-lee: Let's remove this then

```suggestion
    return process.env.ZUPLO_SERVER_URL;
```

PR #2023, opened 2026-02-20
Add env for zuplo gateway url

max-zuplo:


mosch:
@copilot let's rename it to be called `ZUPLO_GATEWAY_SERVICE_URL` also check out how the environment vars are made available in the Zudoku context (check main.tsx) and we could get the env var from the zudoku context since it's already passed into the default query + mutation function. so if it's defined on the context use this if not use the fallback ("https://api.zuploedge.com")

mosch:
@copilot can you make sure ZUPLO_GATEWAY_SERVICE_URL is passed into the context correctly? checkout `main.tsx`

mosch:
@cursoragent can you make sure ZUPLO_GATEWAY_SERVICE_URL is passed into the context correctly? i think its missing in `main.tsx`

PR #2057, opened 2026-02-24
Fix optional query params staying inactive when filled

dan-lee:
Optional schema-defined query params in the playground stayed inactive even after the user entered a value, so they were never included in the request URL. Now params with a value are also set to active.


lcampos (commented):
Reviewed and approved

PR #2100, opened 2026-03-04
chore: add GitHub App token generation and trigger zudoku builder rebuild in release workflow

vazexqi:
This is needed since the release of the https://www.npmjs.com/package/@zuplo/zudoku-plugin-monetization is not always tied to the release of zudoku.

So if you want to have a new version of the plugin in working-copy, we need to trigger the workflow.

See https://github.com/zuplo/core/pull/7438


dan-lee:
Reviewed and approved

PR #2139, opened 2026-03-10
show unique features only

mosch:


packages/plugin-zuplo-monetization/src/utils/formatPrice.ts
```diff
@@ -2,5 +2,6 @@ export const formatPrice = (amount: number, currency?: string) =>
   new Intl.NumberFormat(undefined, {
     style: "currency",
     currency: currency || "USD",
-    minimumFractionDigits: 0,
+    minimumFractionDigits: Number.isInteger(amount) ? 0 : 2,
+    maximumFractionDigits: 2,
```
dan-lee: Not sure about that. What about values like `0.005`?

PR #2186, opened 2026-03-20
feat: add options to hide AI assistant items in schema download menu

mosch:
## Summary
- Adds `useInClaude` and `useInChatGPT` boolean options to the `schemaDownload` configuration
- Both default to `true` (shown) for backward compatibility
- Setting either to `false` hides the corresponding menu item from the download schema dropdown

This is useful when the API schema requires authentication, since Claude and ChatGPT cannot access authenticated URLs.

### Example

```ts
schemaDownload: {
  enabled: true,
  useInClaude: false,
  useInChatGPT: false,
}
```

## Changes
- `interfaces.ts` — Added `useInClaude` and `useInChatGPT` to the `schemaDownload` type
- `ZudokuConfig.ts` — Added both fields to the Zod validation schema
- `ApiHeader.tsx` — Passes `schemaDownload` config to `DownloadSchemaButton`
- `DownloadSchemaButton.tsx` — Conditionally renders Claude/ChatGPT items
- `api-reference.md` — Documented the new options

### DIFF
--- a/docs/pages/docs/configuration/ai-assistants.md
+++ b/docs/pages/docs/configuration/ai-assistants.md
@@ -1,0 +1,65 @@
+---
+title: AI Assistants
+sidebar_icon: bot
+description:
+  Configure which AI assistant integrations appear in dropdown menus across your Zudoku
+  documentation site, including built-in presets and custom providers.
+---
+
+By default, Zudoku shows "Use in Claude" and "Use in ChatGPT" options in dropdown menus on both
+[API reference](./api-reference.md) and [documentation](./docs.md) pages. You can customize this
+behavior using the top-level `aiAssistants` configuration.
+
+## Disable AI Assistants
+
+To remove all AI assistant options:
+
+```ts title=zudoku.config.ts
+const config = {
+  aiAssistants: false,
+  // ...
+};
+```
+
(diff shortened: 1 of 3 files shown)


dan-lee:
I think this should be more flexible than per-provider booleans. Instead of `useInClaude`/`useInChatGPT`, what about a top-level `aiAssistants` config that works for both API reference and docs pages?

```ts
aiAssistants: [
  "claude", // preset
  "chatgpt",
  "copilot",
  "cursor",
  {
    label: "Open in MyAI",
    // simple string with placeholder
    url: "https://myai.com/?context={pageUrl}",
    // or callback for full control
    url: ({ pageUrl, type }) => {
      if (type === "openapi") {
        return `https://myai.com/?q=${encodeURIComponent("Explain this API: " + pageUrl)}`;
      }
      return `https://myai.com/?q=${encodeURIComponent("Explain this page: " + pageUrl)}`;
    },
  },
]
// or disable entirely:
aiAssistants: false
```

The callback receives `{ pageUrl: string, type: "docs" | "openapi" }` so users can customize behavior per context. Built-in presets define their own URL, icon, and label.

This also covers docs pages which have the same Claude/ChatGPT buttons but aren't addressed here.

PR #2194, opened 2026-03-24
fix: preserve $ref sibling properties in nested schemas

### LINKED ISSUE #2191: Fields with `$ref` and OAS3.1-style `description` override do not show description when nested
For an OpenAPI 3.1 schema with the following `components/schemas`:

```yaml
components:
  schemas:
    Child:
      type: string
    ObjectContainingRef:
      type: object
      properties:
        child:
          $ref: '#/components/schemas/Child'
          description: Description for ObjectContainingRef.child
    Parent:
      type: object
      properties:
        childObject:
          type: object
          properties:
            child:
              $ref: '#/components/schemas/Child'
              description: Inline description for childObject.child
        childObjectRef:
          $ref: '#/components/schemas/ObjectContainingRef'
        childObjectArray:
          type: array
          items:
            type: object
            properties:
              child:
                $ref: '#/components/schemas/Child'
                description: Inline description for childObjectArray.[].child
        childObjectArrayRef:
          type: array
          items:
            $ref: '#/components/schemas/ObjectContainingRef'
```

the overridden descriptions for `$ref` fields in the nested objects that are inlined are included in the generated documentation, but those of identical fields in the schemas objects that are defined by `$ref` are not. See below:

Image

The descriptions are correctly included for the `$ref`'ed object's top level representation, but they should also be included when the `$ref`'ed object is nested within another schema definition (i.e. the fields in the screenshot above with a blank description should show the description "Description for ObjectContainingRef.child").

mosch:
## Summary

- Fixes #2191 — fields with `$ref` and OAS 3.1 `description` overrides lost their description when the containing schema was itself referenced via `$ref`
- **File pipeline** (`schema-codegen.ts`): Pass 1 populated `__refMap` entries with `ignoreSiblings=true`, stripping sibling properties from nested `$ref` nodes. Fixed by declaring merged placeholders upfront and using sibling-aware markers in Pass 1.
- **URL pipeline** (`dereference/index.ts`): `$ref` resolution replaced the entire node with the resolved value, losing all sibling properties. Fixed by collecting siblings before resolution and merging them into the result.

## Test plan

Tests were written first against the unmodified code to verify no regressions:

| | Before fix | After fix |
|---|---|---|
| Existing behaviour tests (24) | 24 pass | 24 pass |
| Bug-specific tests (7) | 7 fail | 7 pass |
| Full suite (581) | — | 581 pass |

**New dereference tests** (12 total):
- Existing behaviour: basic resolution, nested refs, arrays, array items, circular refs, custom resolvers, non-ref properties
- Bug tests: sibling description, sibling summary, sibling override, nested `$ref`'d schemas (issue scenario), multiple siblings

**New codegen tests** (8 new, 19 total):
- Existing behaviour: chained refs, object identity, deep paths, no-refs, mixed sibling/non-sibling, circular refs with siblings
- Bug tests: nested referenced schemas with siblings (issue scenario), multi-level nesting

### DIFF
--- a/packages/zudoku/src/lib/oas/parser/dereference/index.ts
+++ b/packages/zudoku/src/lib/oas/parser/dereference/index.ts
@@ -41,12 +41,27 @@ export const dereference = async (
         if ("$ref" in current && typeof current.$ref === "string") {
           // Store the ref path before resolving
           current.__$ref = current.$ref;
+
+          // Collect OAS 3.1 sibling properties to merge after resolution
+          const { $ref: _, __$ref: __, ...siblings } = current;
+          const hasSiblings = Object.keys(siblings).length > 0;
+
           for (const resolver of resolvers) {
             const resolved = await resolver(current.$ref);
-            if (resolved) return await resolve(resolved, path);
+            if (resolved) {
+              const result = await resolve(resolved, path);
+              if (hasSiblings && isIndexableObject(result)) {
+                return { ...result, ...siblings };
+              }
+              return result;
+            }
           }
           const resolved = await resolveLocalRef(cloned, current.$ref);
-          return await resolve(resolved, path);
+          const result = await resolve(resolved, path);
+          if (hasSiblings && isIndexableObject(result)) {
+            return { ...result, ...siblings };
+          }
+          return result;
         }
 
         for (const key in current) {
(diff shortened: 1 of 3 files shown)


dan-lee:
Already addressed in #2192

PR #2239, opened 2026-03-30
fix: don't delete source files when cleaning up generated markdown

henriktre:
## Bug

When `llmsTxt` or `llmsTxtFull` is enabled, the build writes
`node_modules/.zudoku/markdown-info.json` containing the path of each
processed file — but it stores the **source** path (e.g. `pages/documentation.mdx`),
not the generated output path in `dist/`.

After the build, `prerender.ts` cleans up the generated `.md` files when
`publishMarkdown` is `false` (the default). Because `info.filePath` pointed at
the source files, this cleanup deleted the original docs from `pages/` instead.

## Fix

Reconstruct the output path from `info.routePath` and `distDir` — the same
formula already used in `plugin-markdown-export.ts` when writing the files:

```ts
const segments = info.routePath === "/"
  ? ["index"]
  : info.routePath.split("/").filter(Boolean);
const outputPath = `${path.join(distDir, ...segments)}.md`;
rm(outputPath)
```

Only the generated `dist/` files are now removed. Source files are untouched.

## Reproduce

In `examples/cosmo-cargo/zudoku.config.tsx`, change `docs` to:

```ts
docs: {
  llms: { llmsTxt: true },
},
```

Then build:

```bash
pnpm --filter cosmo-cargo build
```

**Before:** `pages/` contains `documentation.mdx`, `global.md`, `intergalactic.mdx`, `interstellar.mdx`, `member-benefits.mdx`, `premium-fleet.mdx`, `quantum-express.mdx`, `shipping-process.mdx`, `tracking.mdx`

**After:** All `.md`/`.mdx` source files are deleted. Run `git restore examples/cosmo-cargo/pages/` to recover them.

### DIFF
--- a/packages/zudoku/src/lib/authentication/components/OAuthErrorPage.tsx
+++ b/packages/zudoku/src/lib/authentication/components/OAuthErrorPage.tsx
@@ -13,19 +13,17 @@ const errorDetailsMap: Record<string, { message: string }> = {
       "The authentication request was invalid. Please try signing in again.",
   },
   unauthorized_client: {
-    message:
-      "This application is not authorized to access your account. Please contact support.",
+    message: "This application is not authorized to access your account.",
   },
   access_denied: {
     message:
       "You denied access to this application. To continue, please sign in and grant access.",
   },
   unsupported_response_type: {
-    message:
-      "The authentication method is not supported. Please contact support.",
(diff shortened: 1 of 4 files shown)


mosch:
@henriktre thanks for your PR!

PR #2242, opened 2026-03-31
Fix on save reformat issues

lcampos:
Enforce format-on-save with the project’s formatter so manual saves stop causing fmt:check failures. Fixes build issues like this https://github.com/zuplo/zudoku/actions/runs/23777033166/job/69281107932


dan-lee:
Reviewed and approved

PR #2252, opened 2026-03-31
Update switch plan flow

lcampos:
This PR does some cleanup so we can re-use the code for calling `purchase-details` endpoint from the`SubscriptionChangeConfirmPage` like we do on `CheckoutConfirmPage`. It also streamlines the flow so we only get one confirmation page when switching plans.
With these changes we now surface VAT/Tax information to the end-user when the project is configured to do so.


dan-lee:
Reviewed and approved

PR #2256, opened 2026-04-01
feat: support robots metadata config

atissedrescaleway:
Hey ! 

We needed to prevent a specific version of our Zudoku app from being indexed, and I noticed the `metadata` config did not support a [robots directive](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/robots).

  This PR adds support for that, so users can control whether their app should be indexed.

## Summary

  Add support for `metadata.robots` in the Zudoku config so users can set robots directives like `noindex, nofollow`.

  ## Changes

  - add `metadata.robots` to the config schema and types
  - render `<meta name="robots" ...>` from metadata config
  - add test for config validation

  ## How to test

  - set `metadata.robots = "noindex, nofollow"` in config
  - verify the page head includes `<meta name="robots" content="noindex, nofollow">`
  - run:
    - `pnpm exec vitest run packages/zudoku/src/config/validators/ZudokuConfig.test.ts`

### DIFF
--- a/packages/zudoku/src/config/validators/ZudokuConfig.test.ts
+++ b/packages/zudoku/src/config/validators/ZudokuConfig.test.ts
@@ -256,6 +256,18 @@ describe("validateConfig", () => {
     expect(mockConsoleLog).not.toHaveBeenCalled();
   });
 
+  it('should accept metadata robots values like "noindex, nofollow"', () => {
+    const config = {
+      metadata: {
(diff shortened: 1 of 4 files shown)


mosch:
Hey @atissedrescaleway – thanks for the PR. LGTM!

dan-lee (approved):

PR #2266, opened 2026-04-06
Fix supabase auth documentation

lcampos:
Updated the Supabase auth docs to reflect the correct way to configure it with Zudoku.

### DIFF
--- a/docs/pages/docs/configuration/authentication-supabase.md
+++ b/docs/pages/docs/configuration/authentication-supabase.md
@@ -2,27 +2,99 @@
 title: Supabase Setup
 sidebar_label: Supabase
 description:
-  Learn how to set up Supabase authentication for Zudoku, leveraging Supabase's built-in auth
-  providers for secure documentation access.
+  Learn how to set up Supabase authentication for Zudoku using email and password or OAuth providers
+  for secure documentation access.
 ---
 
 Supabase is an open-source Firebase alternative that provides authentication, database, and storage
 services. This guide shows you how to integrate Supabase authentication with your Zudoku
 documentation site.
 
+You can use **email and password** (Supabase signs users in with an email address and password),
+**OAuth providers** (GitHub, Google, and so on), or both. Follow the section that matches how you
+want users to sign in.
+
 ## Prerequisites
 
 You'll need a Supabase project. If you don't have one,
 [create a free Supabase project](https://supabase.com/dashboard) to get started.
 
-:::info
+## Setup Steps for Email and Password
 
-Supabase projects often have **email and password** enabled in the dashboard by default. Zudoku's
-Supabase integration is **OAuth only**: email/password and magic links are not supported.
+Use this path when users should sign up and sign in with an email address and password managed by
+Supabase.
 
-:::
+<Stepper>
 
-## Setup Steps
+<a id="configure-email-sign-in"></a>
+
+1. **Enable email authentication in Supabase**
+
+   In your [Supabase Dashboard](https://supabase.com/dashboard):
+   - Select the project you are going to use
+   - Go to **Authentication** → **Configuration** → **Sign In / Providers**
+   - Under **Email**, ensure **Email sign-in** is enabled
+   - Decide whether new users must confirm their email before signing in (see **Confirm email** in
+     the same area). If confirmation is required, Supabase sends a link that returns users to your
+     Zudoku portal (see [Authentication Routes](#authentication-routes))
+
+2. **Copy your Supabase URL and publishable key**
+
+   From your Supabase project dashboard:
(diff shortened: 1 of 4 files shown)


docs/pages/docs/configuration/authentication-supabase.md
```diff
@@ -15,67 +15,79 @@ documentation site.
 You'll need a Supabase project. If you don't have one,
 [create a free Supabase project](https://supabase.com/dashboard) to get started.
 
+:::info
+
+Supabase projects often have **email and password** enabled in the dashboard by default. Zudoku's
+Supabase integration is **OAuth only**: email/password and magic links are not supported.
+
```
mosch: This claim is incorrect. Zudoku's Supabase integration **does** support email/password — it's actually the default behavior.

In `packages/zudoku/src/lib/authentication/providers/supabase.tsx`:
- Line 65: `this.enableUsernamePassword = !config.onlyThirdPartyProviders` — email/password is **enabled by default**
- Lines 159-172: `onUsernamePasswordSignIn` calls `this.client.auth.signInWithPassword()` — fully implemented
- Lines 174-210: `onUsernamePasswordSignUp` calls `this.client.auth.signUp()` — fully implemented
- Password reset and update flows are also fully wired up

The UI in `ZudokuAuthUi.tsx` renders email/password fields and a "Forgot password?" link whenever `enableUsernamePassword` is true (lines 210-229).

Magic links are indeed not supported (no `signInWithOtp` call exists), so that part is accurate.

docs/pages/docs/configuration/authentication-supabase.md
```diff
@@ -15,67 +15,79 @@ documentation site.
...
-   - Go to **Settings** → **Data API**
-   - Copy your **Project URL** (looks like `https://your-project.supabase.co`)
-   - Go to **Settings** → **API Keys**
-   - Copy your **anon public** API key
+   - Go to **Project Settings** → **Integrations** → **Data API**
+   - Copy your **API URL** (looks like `https://your-project-id.supabase.co`)
+   - Go to **Configuration** → **API Keys**
+   - Copy your **publishable** key (looks like `sb_publishable_...`)
 
 3. **Configure Zudoku**
```
mosch, line 119: Adding `onlyThirdPartyProviders: true` to the primary config example implies it's required for basic setup. Since email/password is a supported feature (enabled by default), this should only appear in the "Configuration Options" section where it's documented as optional. The basic example should show the simplest working config without disabling features.

docs/pages/docs/configuration/authentication-supabase.md
```diff
@@ -205,19 +228,23 @@ CREATE TRIGGER on_auth_user_created
 1. **"At least one provider must be provided" Error**: You must configure the `providers` option in
    your authentication configuration with at least one authentication provider in the array.
 
-2. **Invalid API Key**: Ensure you're using the `anon public` key, not the `service_role` key from
-   your Supabase project settings.
+2. **Email/password fields on sign-in**: Zudoku's Supabase integration is OAuth-only. Set
+   `onlyThirdPartyProviders: true` so the sign-in and sign-up screens do not show email/password
+   fields (those flows are not supported with this auth type).
```
mosch: Same issue — this says email/password "flows are not supported with this auth type", but they are fully implemented. The email/password fields appearing is the expected default behavior, not a problem to troubleshoot.

mosch (changes requested):
The dependency removal (`@supabase/auth-ui-react`, `@supabase/auth-ui-shared`, and their workspace overrides) is correct and safe — these packages are not imported anywhere in the codebase.

However, the documentation changes have two significant issues:

1. **Incorrect claim that email/password is not supported.** The Supabase provider fully implements email/password auth (sign in, sign up, password reset, password update). It's enabled by default when `onlyThirdPartyProviders` is not set. The info box and troubleshooting section incorrectly tell users this doesn't work.

2. **Docs still reference removed packages.** Lines 76 and 250 still tell users to install `@supabase/auth-ui-react` and `@supabase/auth-ui-shared`, contradicting the dependency removal in this same PR.

The redirect URL setup guidance and dashboard navigation updates look good.

docs/pages/docs/configuration/authentication-supabase.md
Lewis65: Hello! I think this misunderstanding comes from slightly unclear wording in https://zudoku.dev/docs/configuration/authentication#supabase:

Image
The example and wording make it sound like you must supply a provider, can only use one, and it must be one from that list. LLMs seem to get tripped up by this as well.

This is outlined much better in https://zudoku.dev/docs/configuration/authentication-supabase#supported-providers though :)

vazexqi:
Reviewed and Approved

PR #2268, opened 2026-04-07
Surface auth provider on zuplo build output

lcampos:
Surface the authentication provider type on zuplo build output file

### DIFF
--- a/packages/zudoku/src/vite/build.ts
+++ b/packages/zudoku/src/vite/build.ts
@@ -167,9 +167,13 @@ const runPrerender = async (options: PrerenderOptions) => {
     });
 
     if (ZuploEnv.isZuplo && issuer) {
+      const provider = config.authentication?.type
+        ? config.authentication?.type
+        : "none";
+
       await writeFile(
         path.join(dir, DIST_DIR, ".output/zuplo.json"),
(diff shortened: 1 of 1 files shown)


packages/zudoku/src/vite/build.ts
```diff
@@ -167,9 +167,13 @@ const runPrerender = async (options: PrerenderOptions) => {
     });
 
     if (ZuploEnv.isZuplo && issuer) {
+      const provider = config.authentication?.type
+        ? config.authentication?.type
+        : "none";
```
mosch:
```suggestion
      const provider = config.authentication?.type ?? "none";
```

packages/zudoku/src/vite/build.ts
```diff
@@ -167,9 +167,11 @@ const runPrerender = async (options: PrerenderOptions) => {
     });
 
     if (ZuploEnv.isZuplo && issuer) {
+      const provider = config.authentication?.type ?? "none";
```
mosch:
```suggestion
      const provider = config.authentication?.type;
```

mosch: We should not default to `none` but use `undefined`

mosch (approved):

Issue #2261, opened 2026-04-04
WCAG 4.1.2 violations: Icon-only buttons missing accessible names


aherrera-gropp-tech:
## Description
Several components have ARIA 4.1.2 "Name, Role, Value" violations where icon-only buttons lack accessible names. The lucide-react icons automatically add `aria-hidden="true"`, leaving the parent buttons with no accessible content for screen readers.

### 1. NavigationCategory.tsx (lines 40-59)
The collapsible category toggle button only contains a `ChevronRightIcon`:
```tsx
<Button variant="ghost" size="icon" className="...">
  <ChevronRightIcon size={16} className="..." />
</Button>
```
**Fix:** Add `aria-label` and `aria-expanded`:
```tsx
<Button 
  variant="ghost" 
  size="icon" 
  aria-label={open ? `Collapse ${category.label}` : `Expand ${category.label}`}
  aria-expanded={open}
>
```

### 2. MobileTopNavigation.tsx (lines 40-42)
```tsx
<DrawerTrigger className="lg:hidden">
  <MenuIcon size={22} />
</DrawerTrigger>
```
**Fix:** Add `aria-label="Open navigation menu"`

### 3. Secret.tsx (lines 99-120)
Reveal and copy buttons only contain icons:
```tsx
<Button variant="ghost" size="icon">
  {revealed ? <EyeOffIcon /> : <EyeIcon />}
</Button>
<Button variant="ghost" size="icon">
  {isCopied ? <CheckIcon /> : <CopyIcon />}
</Button>
```
**Fix:** Add `aria-label` to both buttons

### 4. Value.tsx (lines 30-39)
Same copy button issue as Secret.tsx

## Impact
Screen reader users cannot determine the purpose of these buttons, making navigation and interaction inaccessible.

## WCAG Reference
[WCAG 4.1.2 Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)

PR #2273, opened 2026-04-08
fix(a11y): add accessible name to mobile navigation menu trigger :bug:

swapnilgm:
## Summary

- Adds `aria-label="Open navigation menu"` to the mobile header `DrawerTrigger` in `MobileTopNavigation.tsx`, so the control has an accessible name when it only contains a Lucide icon (icons use `aria-hidden`, which left the button unnamed for assistive tech).

This implements the **second** item described in https://github.com/zuplo/zudoku/issues/2261 (WCAG 4.1.2 Name, Role, Value — icon-only buttons missing accessible names).

Relates to #2261.

## How to test

1. Run the Cosmo Cargo sample: `nx run cosmo-cargo:dev`.
2. Narrow the viewport until the menu control appears (below the `lg` breakpoint).
3. **Successful:** The trigger exposes an accessible name (e.g. "Open navigation menu") in the accessibility tree; axe and similar checks should not report an unnamed button.
4. **Unsuccessful:** No accessible name when the menu icon is the only content.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -121,8 +121,8 @@ jobs:
         env:
           ZUDOKU_RELEASE_TOKEN: ${{ steps.zudoku-release-token.outputs.token }}
         run: |
-          auth_header=$(printf 'x-access-token:%s' "$ZUDOKU_RELEASE_TOKEN" | base64 -w0)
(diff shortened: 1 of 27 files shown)


dan-lee:
Thanks for your PR! I think there was another PR that already fixed this, so there are no visible files changed here

PR #2279, opened 2026-04-09
feat: spread userinfo response into UserProfile for custom OIDC claims

Newspicel:
The refreshUserProfile method only mapped 5 hardcoded fields from the userinfo response, discarding custom OIDC claims like Keycloak's resource_access or Auth0 custom claims. This made role-based route protection impossible without manually decoding the JWT access token.

Changes:
- Spread full userinfo response into profile in refreshUserProfile to match existing handleCallback behavior
- Widen UserProfile index signature to accept unknown values so nested objects like resource_access are representable
- Add tests verifying custom claims survive in both handleCallback and refreshUserProfile


Newspicel:
Any possibility to review this or add something similar? @dan-lee @mosch

mosch:
@Newspicel thank you for the ping! We'll look into this now.

Issue #2285, opened 2026-04-10
`wrap-anywhere` in tables is messing up the layout


jesi-rgb:
As you can see in the screenshots, `wrap-anywhere` at the `code` level makes it so that, indeed, text is wrapped anywhere and looks quite off.

All these examples can be seen in production: https://unikraft.com/docs/platform/instances#instance-states

This is as simple as substituting this property for `text-nowrap` or `whitespace-nowrap`. In my testing I found no difference, though in other examples we might see a slight shift. 

In any case, I think these are best kept in one line. We could consider making the whole table horizontally scroll if there was demand for "wider" tables, but in any case, I feel like this little change makes the document consumption much better and easier to parse. Which, I feel, is the ultimate goal a piece of documentation should pursue.

<img width="1029" height="513" alt="Image" src="https://github.com/user-attachments/assets/9755ef8f-48dc-4d72-a534-0edd1045ab31" />

<img width="1131" height="682" alt="Image" src="https://github.com/user-attachments/assets/030e8838-8a51-4a1a-973e-852d1206301f" />

<img width="1841" height="1265" alt="Image" src="https://github.com/user-attachments/assets/f709556a-8665-4f84-b6b8-ed2cdd9dcc7c" />

mosch:
Fixed in #2286

PR #2286, opened 2026-04-10
fix: Do not `wrap-anywhere` if code element belongs to a table

jesi-rgb:
If `code` belongs to a table, let's not wrap it:

image

In normal text, it still wraps as expected: 
image

Aims to solve #2285


jesi-rgb:
I created the new example page to showcase how it works, but happy to undo/remove those changes if necessary!

mosch:
Great increment, thank you!

mosch (approved):

PR #2313, opened 2026-04-14
fix: validate config schema and show clear error messages

### LINKED ISSUE #894: Improve error message for invalid config file

mosch:
When config loading fails (bad imports, syntax errors), show "Invalid Zudoku configuration at <path>" with the underlying error details instead of the misleading "no config file found" message. Also check for missing default export. When falling back to cached config in dev mode, log the error instead of silently swallowing it.

Closes #894

### DIFF
--- a/packages/zudoku/src/config/loader.ts
+++ b/packages/zudoku/src/config/loader.ts
@@ -66,27 +66,44 @@ async function loadZudokuConfigWithMeta(
 ): Promise<ConfigWithMeta> {
   const configPath = await getConfigFilePath(rootDir);
 
-  const { module, dependencies } = await runnerImport<{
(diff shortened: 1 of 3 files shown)


dan-lee (approved):

mosch:
Closing in favor of #2540, which rebases this onto current `main` (this branch was merge-conflicting and built on the pre-`globalThis` loader), drops the "check the logs above" wording that #894 explicitly asked to remove, and adds the `loader.ts` test coverage that was missing here. Original issue: #894.

---

PR #2336, opened 2026-04-16
Fix pagefind `ENOBUFS` by batching indexing after prerender

dan-lee:
During prerendering, `pagefindIndex.addHTMLFile()` was called inside the `Promise.all` worker loop. With large route counts and multiple workers completing concurrently, the pagefind child process IPC pipe overflows with ENOBUFS.

Moves pagefind indexing to a separate batched phase (40 concurrent) that runs after all prerendering completes. Also adds timing to the pagefind log message.

### DIFF
--- a/packages/zudoku/src/vite/prerender/prerender.ts
+++ b/packages/zudoku/src/vite/prerender/prerender.ts
@@ -137,13 +137,6 @@ export const prerender = async ({
     paths.map(async (urlPath) => {
       const result = await pool.run({ urlPath } satisfies WorkerData);
 
-      if (result.statusCode < 400) {
-        await pagefindIndex?.addHTMLFile({
-          url: urlPath,
-          content: result.html,
-        });
-      }
-
       completedCount++;
 
       if (isTTY()) {
@@ -163,10 +156,6 @@ export const prerender = async ({
     }),
   );
 
-  const pagefindWriteResult = await pagefindIndex?.writeFiles({
-    outputPath: path.join(distDir, "pagefind"),
-  });
-
   const seconds = ((performance.now() - start) / 1000).toFixed(1);
 
   const message = `✓ finished prerendering ${paths.length} routes in ${seconds} seconds using ${maxThreads} workers`;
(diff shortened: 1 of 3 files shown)


packages/zudoku/src/vite/prerender/prerender.ts
```diff
@@ -176,10 +165,44 @@ export const prerender = async ({
...
+      const batch = pagesToIndex.slice(i, i + BATCH_SIZE);
+      await Promise.all(
+        batch.map((r) =>
+          pagefindIndex.addHTMLFile({ url: r.url, content: r.html }),
+        ),
+      );
+      if (isTTY()) {
+        const count = Math.min(i + BATCH_SIZE, pagesToIndex.length);
+        writeLine(
+          `pagefind indexing (${count}/${pagesToIndex.length}) ${colors.dim(batch.at(-1)?.url)}`,
```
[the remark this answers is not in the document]
dan-lee: meh it's a non-issue

mosch (approved):

packages/zudoku/src/vite/prerender/prerender.ts
```diff
@@ -87,8 +88,84 @@ export const prerender = async ({
       paths.push(joinUrl(r.from));
     }
   }
-  const maxThreads =
-    buildConfig?.prerender?.workers ?? Math.floor(os.cpus().length * 0.8);
```
vazexqi, line 91: So with large CPUs where you have 32 cores, what's going to happen is they're going to end up with close to 25 workers.

That is too much because what's gonna happen is that you can spawn a lot of workers. All of them are gonna take in all the memory, and they're gonna exhaust the memory, so instead of a CPU-bound problem, it becomes a memory-bound problem.

packages/zudoku/src/vite/prerender/prerender.ts
```diff
@@ -87,8 +88,84 @@ export const prerender = async ({
...
+  const availableForWorkersMb = totalMemMb - reservedMb;
+
+  // Step 3: Per-worker heap limit. This overrides NODE_OPTIONS and
+  // forces V8 to GC aggressively within this budget.
+  const PER_WORKER_HEAP_LIMIT_MB = 4096;
+
+  // Step 4: Worker count — the lesser of memory-based, CPU-based, and
+  // hard cap. Beyond 8 workers the gains are marginal and memory
+  // pressure dominates.
+  const MAX_WORKERS = 8;
```
vazexqi: No matter what, we always kept this at a maximum of eight workers. 
So even if we do a lot of complex calculations of the RAM and CPU and everything else, we just cap it at 8 workers because going beyond this is not going to render this. We're not doing crazy rendering with graphics. 8 workers working in parallel is more than sufficient.

packages/zudoku/src/vite/prerender/prerender.ts
```diff
@@ -87,8 +88,84 @@ export const prerender = async ({
...
+  //   14 workers → OOM at ~1159/2390 routes
+  //    7 workers → SUCCESS (2390 routes in 99s)
+  //
+  // Users can override via `prerender.workers` in their build config.
+  // ────────────────────────────────────────────────────────────────────
+
+  // Step 1: Detect actual memory. In containers, os.totalmem() may
+  // report the host's memory, not the container's cgroup limit.
+  const osTotalMb = Math.floor(os.totalmem() / (1024 * 1024));
+  const cgroupMemMb = getContainerMemoryLimitMb();
```
vazexqi: This is just an extra step to be really careful when we're running inside containers, for instance like Kubernetes and stuff, so that we know what the actual CPU count is as well, so that we don't inflate this wrongly.

packages/zudoku/src/vite/prerender/prerender.ts
```diff
@@ -87,8 +88,84 @@ export const prerender = async ({
...
+  const defaultWorkers = Math.min(
+    memBasedWorkers,
+    cpuBasedWorkers,
+    MAX_WORKERS,
+  );
+  const maxThreads = buildConfig?.prerender?.workers ?? defaultWorkers;
+
+  // Step 5: Per-worker resource limits. Cap at PER_WORKER_HEAP_LIMIT_MB
+  // or the fair share of available memory, whichever is smaller.
+  const maxOldGenerationSizeMb = Math.min(
```
vazexqi: It's done to override anything that we put at the NODE_OPTIONS.  See https://github.com/zuplo/core/pull/7623
We set this a bit too generously and a bit erroneously to give it a maximum of 56 GB. That gets inherited into each worker thread, so that's also compounding this.

packages/zudoku/src/vite/prerender/prerender.ts
```diff
@@ -176,10 +246,43 @@ export const prerender = async ({
   } else {
     logger.info(colors.blue(message));
   }
-  if (pagefindWriteResult?.outputPath) {
-    logger.info(
-      colors.blue(`✓ pagefind index built: ${pagefindWriteResult.outputPath}`),
+
+  if (pagefindIndex) {
```
vazexqi, line 175: These are @dan-lee's changes.

PR #2343, opened 2026-04-17
Fix: show monthly tax amount

lcampos:


max-zuplo (commented):
Reviewed and approved

PR #2353, opened 2026-04-20
SSR auth: server-side token verification + logged-out signal

dan-lee:
Stacked on #2320. Two commits:

1. **Server-verified tokens.** The session endpoint no longer trusts a client-submitted profile — each provider implements `verifyAccessToken` and the server derives the profile from the verified token. Cookie lifetimes are bounded by the token's `exp`. cookie-sync is now provider-agnostic (reads tokens from `providerData`), which keeps provider internals off the critical path. Also emits an authoritative SSR "logged out" signal (`profile: null`) so logged-out users render the Login button on first paint instead of flashing a skeleton.

2. **Azure B2C** opts into SSR auth via OIDC discovery + jose. Auth0 already covered by openid inheritance. Firebase deferred (idToken rotation in the SDK doesn't map cleanly without more work).

### DIFF
--- a/packages/zudoku/src/app/entry.server.tsx
+++ b/packages/zudoku/src/app/entry.server.tsx
@@ -10,9 +10,10 @@ import {
   type RouteObject,
 } from "react-router";
 import "vite/modulepreload-polyfill";
+import { configuredAuthProvider } from "virtual:zudoku-auth";
 import config from "virtual:zudoku-config";
 import { parseCookies } from "../lib/authentication/cookies.js";
-import { sessionHandler } from "../lib/authentication/session-handler.js";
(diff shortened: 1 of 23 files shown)


packages/zudoku/src/app/entry.server.tsx
```diff
@@ -72,8 +72,10 @@ export const handleRequest = async ({
   }
 
   const { accessToken, profile } = parseCookies(request);
```
mosch, line 74: why is the profile in cookie?

mosch (commented):
Reviewed & Approved

dan-lee:
Reviewed and approved

PR #2359, opened 2026-04-20
fix: prefix basePath on favicon href and clarify GitHub Pages deploy

### LINKED ISSUE #2358: `config.basePath` does not work as is with GitHub pages
We are trying to use `config.basePath` to prepare our docs to be served using GitHub pages, but are facing few problems:

1. If `config.basePath` is set to `/your-repo`, the `api` and `assets` directories as well as `api.html` and `index.html` are inside `your-repo` directory. This causes bunch of 404 responses as GitHub pages is looking for these files from the root of the directory used for the pages deployment.

    As a workaround, we had to add following to our build script:

    ```sh
    if [ -n "$ZUDOKU_PUBLIC_BASE_PATH" ]; then
      mv "docs/$ZUDOKU_PUBLIC_BASE_PATH/"* docs/
      rmdir "docs/$ZUDOKU_PUBLIC_BASE_PATH"
    fi
    ```

2. Favicon path does not contain the `config.basePath`. As a workadound, we had to add following to our Zudoku config:

    ```ts
    const favicon = basePath ? `${basePath}/favicon.svg` : "/favicon.svg";
    ```

dan-lee:
Fixes #2358.

Two issues for users setting `basePath`:

1. The favicon `<link rel="icon">` skipped the basePath prefix, so with `basePath: "/your-repo"` and a root-relative `favicon: "/favicon.svg"` the href went out as `/favicon.svg` and 404'd. Absolute URLs pass through untouched.
2. The GitHub Pages deploy docs said to upload `dist/`, but Zudoku writes into `dist/<basePath>/`. Since Pages already serves a project site at `/your-repo/`, uploading the outer `dist/` double-nests everything under `/your-repo/your-repo/` and every asset 404s. The fix is to upload `dist/<basePath>/`.

New `examples/with-base-path` exercises basePath + favicon + schema download + llms.txt, and a CI smoke step asserts the layout so favicon prefixing and the nested dist structure cannot silently regress.


dan-lee:
@claude review

mosch (approved):

PR #2380, opened 2026-04-23
fix: preserve 'ansi' language in CodeTabs

nurof3n:
`ansi` language is a special built-in in shiki that does not appear in `highlighter.getLoadedLanguages()`, even though it is available.

This PR adds support for it, like in normal code blocks. This is useful for adding terminal coloring.


dan-lee (approved):
Thanks for your contribution @nurof3n! LGTM

PR #2384, opened 2026-04-24
Protect SSR-only routes with chunk-level gating

dan-lee:
Builds on `feat/ssr-with-auth`. Gates protected routes end to end so the JS for auth-only pages never ships to anonymous visitors.

Route-shaped dynamic imports get auto-detected at build time and moved into a `_protected/` chunk directory. At runtime those chunks are served from the SSR origin behind an access-token cookie (verifier reused from the auth provider, fail-closed), `_protected/*` bypasses the CDN, and a `findProtectedLeaks` check fails the build if a public entry statically reaches a protected chunk. On the client, `wrapProtectedRoutes` replaces `lazy()` with a no-op for unauth users on protected paths so the 401 from the asset middleware doesn't leave the page half-hydrated.


mosch (commented):
Reviewed & Approved

PR #2398, opened 2026-04-27
Replace login dialog with auto-redirect to auth provider

mosch:
## Summary
Replaced the modal login dialog with automatic redirection to the authentication provider when users attempt to access protected routes. This simplifies the UX by removing the intermediate dialog step and leveraging the auth provider's native sign-up/login flows.

## Key Changes
- **Removed LoginDialog component**: Eliminated the modal dialog that prompted users to login or register
- **Implemented auto-redirect flow**: Added automatic redirection to `auth.signup()` when users are unauthorized, with a loading state ("Redirecting to login…") displayed while the redirect is in progress
- **Preserved URL state**: Updated redirect logic to include search parameters and hash fragments in the `redirectTo` URL, ensuring users return to the exact location they requested
- **Simplified RouteGuard logic**: Removed `useNavigate` hook and dialog state management; replaced with a `useRef`-guarded effect that initiates the redirect once per unauthorized session
- **Updated navigation blocking**: When users attempt to navigate to protected routes, the blocker now triggers the auto-redirect instead of showing a dialog with cancel/login/register buttons
- **Updated documentation**: Revised protected routes docs to reflect the new auto-redirect behavior and removed references to the login dialog

## Implementation Details
- Added `RedirectingToLogin` component to display a simple loading message during the redirect
- Used `redirectInitiatedRef` to prevent duplicate `auth.signup()` calls across re-renders
- Consolidated authorization checks into `isUnauthorizedDestination` variable for clearer logic
- Preserved hash fragments in redirect URLs (previously only pathname and search were included)
- Tests updated to verify auto-redirect behavior and removal of dialog UI

https://claude.ai/code/session_018ptAjFRWx35ioE4Gzk8tGV

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,3 +9,7 @@ See: https://github.com/zuplo/zudoku/releases
 - Merged `topNavigation`, `sidebar` and `customPages` into a single `navigation` configuration.
 - Plugins now require a `path` property instead of `navigationId`.
 - Added migration guide in `docs/pages/guides/navigation-migration.md`.
+- **Breaking:** Protected routes now redirect unauthenticated users straight to the authentication
+  provider instead of showing a "Login to continue" dialog. The `redirectTo` URL also preserves the
+  hash fragment. The Cancel and Register affordances on the dialog are gone; if you have
+  `redirectToAfterSignIn` configured, it still takes precedence over the originally-requested URL.
--- a/docs/pages/docs/configuration/protected-routes.md
+++ b/docs/pages/docs/configuration/protected-routes.md
@@ -38,9 +38,9 @@ authenticated to access these routes.
 }
 ```
 
-When a user tries to access a protected route, a login dialog will appear prompting them to sign in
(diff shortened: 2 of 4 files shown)


dan-lee:
Two things:

1. The code calls `auth.signup()` while the screen says "Redirecting to login…". For returning users (the common case) that drops them on a registration form at the IdP. Most providers let you switch from login → register but not the other way. Should be `login()`.

2. No escape hatch. If `signup()` rejects, the IdP is misconfigured, or a popup is blocked, `redirectInitiatedRef` stays true and the user is wedged on "Redirecting to login…" with no back button and no retry. Needs a `.catch()` that resets the ref, or a fallback after a timeout.

Also not fully sold on dropping the blocker dialog. Its real job wasn't login/register choice, it was "don't yank me off this page." Right now clicking a wrong link in the nav instantly nukes scroll/form state and punts you to the IdP. A confirm step there felt right even if the unauthorized-on-direct-load case goes straight to redirect.

PR #2399, opened 2026-04-27
Enable private plan switch to public plan 

lcampos:
Allow users on private plans to move to public plans


dan-lee:
Reviewed and approved

PR #2404, opened 2026-04-29
Customize sign-up across auth providers

dan-lee:
Register and Sign in went to the same OIDC authorize endpoint, so there was no way to send users to a separate sign-up page or hide Register on invite-only sites.

This change adds two options to every auth provider. `signUp: { url }` redirects Register to a separate URL (absolute goes external, relative navigates in-app). For openid/auth0 there is also `signUp: { authorizationParams }` to add extra params to the authorize URL on sign-up only (e.g. `kc_action=register` for Keycloak, `screen_hint=signup` for custom OIDC). `disableSignUp: true` hides the Register button on the protected-route login dialog and shows an "Invitation required" page on `/signup`. It is purely visual, real enforcement still belongs at the IdP.

### DIFF
--- a/docs/pages/docs/configuration/authentication-auth0.md
+++ b/docs/pages/docs/configuration/authentication-auth0.md
@@ -154,6 +154,28 @@ When the prompt parameter is omitted (empty string), Auth0 will:
 - Silently authenticate the user if they have a valid session
 - Redirect to the login page if no valid session exists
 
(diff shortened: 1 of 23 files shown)


mosch (approved):

dan-lee:
@copilot resolve the merge conflicts in this pull request

PR #2406, opened 2026-04-29
fix: Render common properties in union view

jedevc:
Previously, if a type had a union (via `oneOf`), but also had properties set directly, we were skipping rendering of those common properties. For example of where we were seeing this https://unikraft.com/docs/api/platform/v1/instances#update-instances.

| Before | After |
| :----- | :---- |
| image | image |

I don't really know if this is the desired rendering state, but it's definitely better than what we had before, which was completely empty!

This is possible to reach by having an `allOf` that contains `oneOf`s, imagine something like this:

```yaml
openapi: 3.1.0
info:
  title: Example
  version: 1.0.0
paths:
  /lookup:
    post:
      requestBody:
        content:
          application/json:
            schema:
              allOf:
                - type: object
                  properties:
                    tenant:
                      type: string
                  required: [tenant]
                - oneOf:
                    - type: object
                      properties:
                        name:
                          type: string
                      required: [name]
                    - type: object
                      properties:
                        uuid:
                          type: string
                          format: uuid
                      required: [uuid]
      responses:
        "200":
          description: OK
```

Or even just directly without the `allOf` as:

```yaml
openapi: 3.1.0
info:
  title: Example
  version: 1.0.0
paths:
  /lookup:
    post:
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                tenant:
                  type: string
              required: [tenant]
              oneOf:
                - properties:
                    name:
                      type: string
                  required: [name]
                - properties:
                    uuid:
                      type: string
                      format: uuid
                  required: [uuid]
      responses:
        "200":
          description: OK
```

### DIFF
--- a/packages/zudoku/src/lib/plugins/openapi/schema/UnionView.tsx
+++ b/packages/zudoku/src/lib/plugins/openapi/schema/UnionView.tsx
@@ -1,8 +1,10 @@
-import { useState } from "react";
+import { Fragment, useState } from "react";
 import type { SchemaObject } from "../../../oas/parser/index.js";
 import { Badge } from "../../../ui/Badge.js";
 import { Frame, FramePanel } from "../../../ui/Frame.js";
+import { ItemGroup, ItemSeparator } from "zudoku/ui/Item.js";
 import { cn } from "../../../util/cn.js";
+import { SchemaPropertyItem } from "./SchemaPropertyItem.js";
 import { SchemaView } from "./SchemaView.js";
 import {
   decideExclusivity,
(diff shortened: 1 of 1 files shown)


dan-lee:
Hey @jedevc, thanks for catching this! I had a go at an alternative where `SchemaView` dispatches to `ObjectSchemaView` for schemas with both own properties and a union, so the existing object grouping gets reused. https://github.com/zuplo/zudoku/pull/2411 

I pushed a pre-release if you'd like to try it out and see whether it covers your case. Just install `zudoku@ 0.0.0-fix-union-view-common-properties.ze562a38b` and give it a go!

Btw, a preview of the fix can be found in our [golden example project here](https://zudoku-cosmo-cargo-git-fix-union-view-common-properties.zuplosite.com/catalog/api-cargo-containers/~schemas#paymentmethod).

Lmk if that would work for you ✌️

PR #2407, opened 2026-04-29
Fix cancel subscription for plans with no current billables

max-zuplo:
The monetization backend rejects timing=next_billing_cycle when the active phase has no billable items, which happens for free plans and subscriptions currently in a free-trial phase of a paid plan. Detect this client-side from phase.items[].price and fall back to timing=immediate, with copy variants for free, trial, and paid cases.

image


dan-lee:
Reviewed and approved

PR #2408, opened 2026-04-29
Fix .env.local to make examples work

max-zuplo:


dan-lee (changes requested):
Tested this and nx already auto-loads `.env.local` from the project root before running a target. So `examples/with-zuplo/docs/.env.local` should already be picked up for `nx run with-zuplo:*` without any Vite changes.

Are you running it some other way? If nx's loading isn't doing what we need, fixing it there feels cleaner than having vite reach for `.env.local` itself, especially since this code path also runs on real Zuplo builds.

Docs: https://nx.dev/recipes/tips-n-tricks/define-environment-variables

max-zuplo:
@dan-lee I am not using nx, and not sure why should I start using it vs directly calling pnpm. I am using `pnpm dev` while in examples/zuplo-monetization/docs, nothing works for me without this change.

dan-lee:
@max-zuplo Fair that nx isn't your flow, but it's the documented setup in [CONTRIBUTING.md](https://github.com/zuplo/zudoku/blob/main/CONTRIBUTING.md) (`nx run <example>:dev`), and going through nx makes `.env.local` work without any framework changes.

For `pnpm dev` directly in an example dir `direnv` or `dotenv-cli` (`dotenv -e .env.local -- pnpm dev`) would work.

PR #2421, opened 2026-05-06
feat(auth): propagate allowInsecureRequests for oauth providers

jedevc:
I've been working with a local oauth provider, and trying to get zudoku to integrate with it, which wasn't possible without this option.


dan-lee (approved):
Thanks for your contribution 🚀

PR #2434, opened 2026-05-12
Refine pricing card display in the monetization plugin

mosch:
Started as a port of the headline + body changes from [zuplo/portal#4751](https://github.com/zuplo/portal/pull/4751), then grew into a broader polish of how plan cards render usage-based pricing, tiered breakdowns, and metered quotas — plus a handful of Copilot-flagged edge cases.

## Summary

- **Headline**: Pay-as-you-go and tiered usage-based plans previously rendered as **"Free"** on `PricingCard` because the headline used only the precomputed monthly flat-fee total. They now surface **"Pay as you go" / "Usage-based pricing"** (sized at `text-2xl` so the phrase fits a 200px card on one line). Hybrid plans (flat + usage) classify as `"priced"` — usage on top of the base is communicated by the per-feature tier breakdown, not a separate top-line annotation.
- **PAYG detection requires actual pricing**: a `usage_based` rate card with `price: null` (used in this repo for free, quota-only metered entitlements) or an all-zero tiered schedule no longer flips a plan into PAYG; such plans correctly classify as `"Free"`.
- **Body**: `usage_based` rate cards without an `issueAfterReset` quota were previously skipped entirely by `categorizeRateCards`, leaving the entitlement list empty. They now render as PAYG `QuotaItem` rows with either a unit price (`API Calls — $0.10/unit`) or a tier breakdown.
- **Tiered display consolidation**: three places previously duplicated the same quota info — a `"X / period"` header, a tier breakdown, and a `"+ $X after quota"` overage line. The tier breakdown is now the single source of truth: the period header is hidden when a breakdown is present (the breakdown's `"Up to X: Included"` line conveys the included quota), and the overage line is removed.
- **Tier line format**: flipped to flat-first (`$499`, `$199 + $0.05/unit`); `Included` shown only when both flat and unit are zero.
- **Priced-first-tier detection**: plans whose first tier has a positive flat or unit price (e.g. a `$499` base) route through PAYG since their issued amount is a tier boundary, not free included usage.
- **Single-tier handling**: single-tier `tiered` prices (the only single-tier shape the backend allows is open-ended) use the same flat-then-unit format the multi-tier breakdown uses — `"$flat + $unit/label"` when both are positive, just one part when the other is zero. Previously the unit branch short-circuited and any flat price was discarded. All-zero tiered schedules fall back to the features bucket; on the subscription detail page, priced single-tier items keep `"X / period"` visible since there's no breakdown to fall back on.
- **`isSoftLimit` no longer gates display**: a usage-based card with `isSoftLimit: false` and a unit price was previously dropped from both `quotas` and `features`, leaving the card body empty. Soft/hard limit is a metering enforcement concern — it doesn't change what the card should show — so any usage-based metered card with a positive price now renders as PAYG.
- **Quota period tracks `usagePeriod`**: the `"X / period"` line now prefers `entitlementTemplate.usagePeriod` (the admin UI's "Quota resets every" field) over the rate card or plan billing cadence. A trial phase with `usagePeriod: "P1W"` inside a `P1M`-billed plan now correctly shows `"10,000 / week"` instead of `"10,000 / month"`.
- **Cadence parsing**: `formatDuration` now handles `PT1H` / `PT15M` / `PT30S` (previously fell through to the raw ISO string), matching the coverage in portal's `formatPeriod` lookup.

## Changes

- New `src/utils/formatPlanPrice.ts` returning a discriminated `{ type: "free" | "payg" | "priced" }` label, used by `PricingCard`.
- `formatPlanPrice`: introduced `isPricedUsageRateCard` so PAYG only triggers when a `usage_based` card has a non-zero unit price or a tier with positive flat/unit amounts.
- Extended `categorizeRateCards` with a PAYG branch for `unit` and `tiered` usage-based prices, including a priced-first-tier heuristic, all-zero-tier detection (→ features bucket), and single-tier handling that composes `"$flat + $unit/label"` the same way the multi-tier breakdown does.
- `categorizeRateCards`: `periodFor` now reads `entitlementTemplate.usagePeriod` first (with `rc.billingCadence` → `planBillingCadence` → `"month"` as fallback chain).
- Extended `Quota` type with `isPayg` and `unitPrice`; dropped `overagePrice` since the tier breakdown now carries that info.
- `QuotaItem` and `SubscriptionPlanDetails` unified on a single rule: hide the `"X / period"` line when `tierPrices` has content. `SubscriptionPlanDetails` gates the suppression on actually having a breakdown to fall back on.
- `SubscriptionPlanDetails` mirrors the `usagePeriod` precedence by reading `entitlement.usagePeriod?.intervalISO` (subscription-side shape) before falling back to `item.billingCadence` / subscription cadence.
- `formatTieredPriceBreakdown`: removed the `omitIncludedUpToAmount` parameter; the included tier is now rendered as `"Up to X: Included"` directly.
- `formatDuration`, `formatDurationInterval`, `formatDurationAdjective`: added hours/minutes/seconds branches. The adjective form stops at hours (sub-hour adjectives like "per-minute" read awkwardly in possessive contexts).
- Dropped the `"+ usage"` annotation block and the `hasUsage` field from `PlanPriceLabel`'s `"priced"` variant — the per-feature tier breakdown already communicates overage where it applies, so the top-line label was redundant.

## Test plan

- [x] `pnpm vitest run` in `packages/plugin-zuplo-monetization` — 223 tests pass (incl. new tests for PAYG categorization, priced-first-tier routing, tier breakdown formatting, hour/minute/second cadences, hard-limit PAYG rendering, `usagePeriod` precedence, single-tier flat+unit composition, all-free edge cases, and `price: null` vs. positive-price PAYG classification)
- [x] `pnpm fmt:check` and `pnpm biome lint` clean
- [x] `pnpm -F @zuplo/zudoku-plugin-monetization typecheck` clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01BCNPbuHC3ZgKCzQvpPwsti)_


lcampos (commented):
Reviewed and approved

PR #2468, opened 2026-05-19
Tighten PricingTable grid sizing and gap

mosch:
## Summary

- Reduce the minimum card width in the PricingTable grid from `300px` to `200px`.
- Switch the second `minmax` arg from `max-content` to `1fr` so columns fill the row evenly instead of shrinking to content.
- Drop the inter-card `gap` from `6` to `2` for a denser layout.
- Bump `@zuplo/zudoku-plugin-monetization` to `0.0.36-pre.2`.

## Why

The previous `300px` / `max-content` / `gap-6` combination left a lot of whitespace and could collapse into a single column on mid-width viewports. The new values pack more plans per row and keep the columns equal width.

## Test plan

- [ ] Visually verify the PricingTable in a consumer (e.g. monetization plugin demo) at narrow, medium, and wide widths.
- [ ] Confirm `renderCard` overrides still receive the expected `defaultCard`.

### DIFF
--- a/packages/plugin-zuplo-monetization/src/pricing-ui/PricingTable.tsx
+++ b/packages/plugin-zuplo-monetization/src/pricing-ui/PricingTable.tsx
@@ -81,7 +81,7 @@ export const PricingTable = ({
     <>
       <div
         className={cn(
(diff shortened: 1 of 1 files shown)


mosch:
@copilot resolve the merge conflicts in this pull request

dan-lee (approved):

PR #2488, opened 2026-05-20
Fixes to switching plans

lcampos:
- Allow switching to a private plan when subscribed to a private plan
- Allow switching to a new version of the plan currently subscribed to


mosch:
Reviewed and approved

PR #2489, opened 2026-05-20
Fix missing type on plugin checks

lcampos:


vazexqi (approved):
LGTM. This is safe.

@dan-lee @mosch might have better ideas on how to do this but this will unblock the monetization plugin.

PR #2556, opened 2026-06-04
Monetization: per-phase price schedule for multi-phase plans

max-zuplo:
Multi-phase (intro + steady) plans previously showed only the last phase's price as a single headline everywhere, hiding the intro phase's price entirely (e.g. a "$375/month for 3 months, then $750/month" plan rendered as just "$750/month").

- Add getPhasePriceLabel + getPlanPriceSchedule: per-phase price labels and a stacked schedule (rows like "First 3 months / After that"), returned only when phase prices actually differ
- New PlanPriceSchedule component renders the schedule on the pricing card, the switch-plan modal card, and the checkout / plan-change confirmation summaries; single-phase plans render exactly as before
- PlanEntitlements collapses identical per-phase feature lists into one (order-insensitive via new sameEntitlementSet); differing phases keep per-phase sections with the phase's own price in the header
- PlanChangeCard mirrors that rule for its entitlement diff: differing phases get one diff section per phase (vs the current subscription), headed by the shared PlanPhaseHeader
- Remove formatPhaseRampSummary (superseded; produced "First 3 months (3 months), then $750 / month" without the intro price)
- Cover getActivePhase / getSubscriptionPlanView multi-phase ramp selection with time-pinned tests (intro price now, steady price after the window passes)

Public pricing-ui exports are additive only: PlanPriceSchedule, getPlanPriceSchedule, PlanPriceScheduleRow, getPhasePriceLabel.


lcampos (commented):
Reviewed & approved

PR #2568, opened 2026-06-08
docs(agent): drop nx guidance, document changesets workflow

mosch:
## What

Updates the agent guide (`AGENT.md`, which `CLAUDE.md` symlinks to) for the nx → Changesets migration:

- **Drops the stale nx instructions.** The `Dev` bullet, the `Monorepo` architecture bullet, and the cosmo-cargo run command no longer reference `nx run …`. They now describe the pnpm + run-from-source workflow (matching the wording in #2513 so the two don't diverge).
- **Adds a `Changesets (changelogs & releases)` section.** This is the part that was missing everywhere: an actionable instruction telling Claude *where changelog entries go now* — run `pnpm changeset`, commit the generated `.changeset/*.md`, and don't hand-edit `CHANGELOG.md` (it's regenerated at release time). It also notes when to skip a changeset.

## Why

The current guide tells Claude to run `nx run docs:dev` / use "pnpm + nx", which becomes wrong the moment the nx removal lands. And neither the old guide nor #2513 ever tells an agent how to author a changelog entry under the new Changesets flow, so changelogs would silently get missed.

## Note on ordering

This documents the world that **#2513** (`chore: drop nx, adopt changesets and pkg.pr.new`) introduces — on `main` today, nx is still the tooling and `.changeset/` doesn't exist yet. This should land **together with / after #2513**. The nx-line edits here are intentionally identical to #2513's, so if both touch the same lines the merge is a no-op rather than a conflict; the new Changesets section is purely additive.

https://claude.ai/code/session_018aLTRzZvQq4c7GvcH57hHb

---
_Generated by [Claude Code](https://claude.ai/code/session_018aLTRzZvQq4c7GvcH57hHb)_

### DIFF
--- a/AGENT.md
+++ b/AGENT.md
@@ -11,14 +11,29 @@
   - Always use `--write` when running biome lint to fix issues in one command
 - **Test**: `pnpm test` or for single test: `pnpm vitest run path/to/test.spec.ts`
 - **Typecheck**: `pnpm -F zudoku typecheck` to check types for the zudoku package
-- **Dev**: Running example projects with `nx` (e.g., `nx run docs:dev`) will automatically rebuild
-  dependent packages as needed. Don't manually run `pnpm -F zudoku build` repeatedly.
+- **Dev**: The zudoku CLI runs from source via `tsx` when `packages/zudoku/dist/` is absent, so
+  example projects (e.g. `pnpm -F docs dev`, `pnpm -F cosmo-cargo dev`) work without first building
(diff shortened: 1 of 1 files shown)


dan-lee:
Integrated into #2513 in commit 9c8fdf4b6c449db32cb6da98469b7e496eea011e (the Changesets agent doc section). The nx-line edits already overlapped with that PR.

PR #2574, opened 2026-06-10
fix: generate `.md` files when `publishMarkdown` default applies

dan-lee:
The `.md` copy-page routes 404 unless a config explicitly sets `docs.publishMarkdown`, even though it defaults to true. Schema defaults never reach the loaded config (`validateConfig` safeParses but discards the result), so the markdown-export plugin read a raw `undefined` and skipped writing the files, while the client still rendered the Copy page button.

This parses the docs block the same way `plugin-docs` and `prerender` already do, so the default applies and `.md` files generate even when `docs` is omitted. The broader issue of defaults not propagating is tracked separately.

### DIFF
--- a/packages/zudoku/src/lib/plugins/openapi/SchemaInfo.tsx
+++ b/packages/zudoku/src/lib/plugins/openapi/SchemaInfo.tsx
@@ -274,6 +274,7 @@ export const SchemaInfo = ({
     data: { schema },
   } = useSuspenseQuery(query);
   const { title, description } = schema;
+  const sidebarCollapsed = useSidebar((s) => s.isCollapsed);
 
   useWarmupSchema();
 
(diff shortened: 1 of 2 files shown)


mosch (approved):
Reviewed the diff. Correct, well-scoped fix: the markdown-export plugin was reading `config.docs?.publishMarkdown` raw, but schema defaults never reach the loaded config, so the `publishMarkdown: true` default was `undefined` and `.md` files were skipped. Parsing `config.docs ?? {}` through `DocsConfigSchema` so defaults apply is the right fix, and it matches the pattern `plugin-docs`/`prerender` already use. Applied consistently across `buildStart`, `configureServer`, and `closeBundle`, and the trailing `llms` check is simplified to reuse the parsed `llmsConfig`. Single file, no behavior change beyond the intended default. LGTM.

---

PR #2600, opened 2026-06-12
refactor: simplify navigation schema recursion with zod 4 getters

dan-lee:
The navigation schema used zod 3 era workarounds for recursion: `z.lazy`, two whole-schema `ZodType` annotations, and a base/`.extend()` split. Zod 4 getters replace these with a single annotated getter. Items now parse through a discriminated union, so a typo in `type` reports an invalid discriminator instead of errors from all seven union branches.

The `InputNavigationItem` type union stays hand-written since TypeScript can't infer recursion cycles that pass through unions. Adds unit tests for the recursive parsing.

### DIFF
--- a/packages/zudoku/src/config/validators/InputNavigationSchema.test.ts
+++ b/packages/zudoku/src/config/validators/InputNavigationSchema.test.ts
@@ -1,0 +1,109 @@
+import { describe, expect, it } from "vitest";
+import {
+  InputNavigationSchema,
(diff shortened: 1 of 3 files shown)


mosch (approved):
Clean simplification — nice use of Zod 4 getters to drop the lazy/extend workarounds. The discriminated union giving clearer errors is a good bonus. LGTM.

---

PR #2612, opened 2026-06-12
fix(plugin-graphql): fall back to plain introspection query for older GraphQL servers

dan-lee:


mosch (approved):
Nice fix — extracting `fetchIntrospection` and falling back to the plain query for older servers is clean and minimal. The `optimizeDeps` addition and the dead `cn.ts` removal are good housekeeping too. LGTM! ✅

---

PR #2613, opened 2026-06-12
fix: copy page button bottom half not clickable

dan-lee:
The h1 rendered by `Heading` has `relative` baked in, so its transparent full-width box painted above the floated page actions and swallowed clicks on the bottom half of the Copy page button. The button container now gets `relative z-10`, with `isolate` on the page `<header>` so the z-index can't escape and compete with the sticky site header.


mosch (approved):
Clean fix — adding `relative z-10` to the button container and `isolate` on the page wrapper is exactly the right scope for this. LGTM.

---

PR #2615, opened 2026-06-13
fix: prevent LazyMotion renderer crash from undefined motion features

dan-lee:
Fixes a crash some users hit on the stacked sub-navigation: `Cannot destructure property 'renderer' of undefined` thrown inside `LazyMotion`.

The lazily imported `motionFeatures` module re-exported `domAnimation` as `export { domAnimation as default } from "motion/react"`. That pure re-export keeps the chunk's default as a transitive live binding, which can resolve to `undefined` under circular chunk init order, so `LazyMotion` destructured `{ renderer }` off an undefined value. Switched to the `import` plus `export default` form the Motion docs recommend, so the chunk holds a concrete value.


mosch (approved):
Good catch — switching from the re-export to a concrete `export default` is exactly the right fix for this circular init issue. Clean and minimal. LGTM.

---

PR #2622, opened 2026-06-15
build: use tsgo for typechecking

dan-lee:
Switches TypeScript typechecking across the zudoku CLI and plugin packages from `tsc` to `tsgo` (`@typescript/native-preview`), the native TypeScript compiler.

The `typecheck` script in `zudoku`, `plugin-graphql`, and `plugin-zuplo-monetization` now runs `tsgo`, and `typecheck.checker: "tsgo"` is set in zudoku's vitest config so `vitest --typecheck` uses it too. Added `@typescript/native-preview` as a catalog devDependency pinned to `7.0.0-dev.20260610.1`.

The full typecheck toolchain now runs in about 3.5s, down from roughly 9s for the zudoku package alone under `tsc`. The root `vitest run --typecheck` still uses `tsc` for projects without type tests, since only zudoku has them.

### DIFF
--- a/packages/plugin-graphql/package.json
+++ b/packages/plugin-graphql/package.json
@@ -21,7 +21,7 @@
     "build": "tsdown",
     "dev": "tsdown --watch",
     "prepublishOnly": "pnpm typecheck && pnpm build",
-    "typecheck": "tsc --project tsconfig.json"
+    "typecheck": "tsgo --project tsconfig.json"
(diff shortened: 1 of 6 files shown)


mosch (approved):
Nice improvement — 9s → 3.5s is a solid win. Clean, focused change. LGTM. ✓

---

PR #2775, opened 2026-08-24
feat: add option to integrate custom react providers

### LINKED ISSUE #1458: Support Custom React Providers
Today we support custom plugins that can register a script in the `<head>` of a project. For example registering posthog:

```
  plugins: [
    {
      getHead: () => (
        <script>
          {`!function(t,e){var o,n,p,r;e.__SV||(window.po....`}
        </script>
      ),
    },
  ],
``` 

However, this requires using the hosted version of posted. What would be better is if we could use `posthog-js`.

Not exactly sure the best react friendly design here, but something like this:

```
  plugins: [
    {
      getProvider: ({children}) => (
        <PostHogProvider apiKey={import.meta.env.VITE_PUBLIC_POSTHOG_KEY} options={options}>
          {children}
        <PostHogProvider/>
      ),
    },
  ],
```

tusharmotwani1718:
### **Support custom React Providers in plugins**

**Overview**

This draft adds support for plugins to register custom React Providers that wrap the Zudoku application.

Currently, plugins can inject elements into the document <head> through getHead(). This works for integrations that expose a script-based setup, but it is not ideal for libraries that provide a React-specific Provider API.

For example, libraries such as PostHog can be integrated using a React Provider:

<PostHogProvider apiKey={...}>
  {children}
</PostHogProvider>

This change explores allowing Zudoku plugins to provide such React Providers directly.

**Proposed API**

A new ProviderPlugin interface is introduced:

export interface ProviderPlugin {
  getProvider?: (args: {
    children: ReactNode;
  }) => ReactNode;

  providerPriority?: number;
}

It is added to the ZudokuPlugin union.

A plugin can then be defined as:

const myPlugin: ZudokuPlugin = {
  getProvider: ({ children }) => (
    <MyProvider>
      {children}
    </MyProvider>
  ),
};

**Provider composition**

Multiple provider plugins are collected and sorted by providerPriority:

const providerPlugins = (props.plugins ?? [])
  .filter(hasProvider)
  .sort(
    (a, b) =>
      (a.providerPriority ?? 0) -
      (b.providerPriority ?? 0),
  );

They are then composed using reduceRight() so that lower priority values become the outermost providers.

For example:

Provider A: priority 10
Provider B: priority 20

results in:

Provider A
└── Provider B
    └── Zudoku

The priority mechanism is currently included mainly to explore how multiple provider integrations should be composed.

**Provider placement**

The resulting provider tree is rendered inside the existing ZudokuProvider hierarchy:

<ZudokuProvider context={zudokuContext}>
  <RouterEventsEmitter />

  <SlotProvider>
    <MDXProvider>
      <ThemeProvider>
        <ViewportAnchorProvider>
          {app}
        </ViewportAnchorProvider>
      </ThemeProvider>
    </MDXProvider>
  </SlotProvider>
</ZudokuProvider>

This means custom providers can potentially consume Zudoku's React context if required.

### Open questions / feedback requested

This is currently a draft because there are a few design questions where I'm not sure what would fit Zudoku's existing plugin architecture best.

**1. Should ProviderPlugin be a separate plugin type?**

Currently it is defined separately:

export interface ProviderPlugin {
  getProvider?: ...
  providerPriority?: number;
}

and added to ZudokuPlugin.

I'm unsure whether this should instead be part of CommonPlugin, for example:

export interface CommonPlugin {
  initialize?: ...
  getHead?: ...
  getMdxComponents?: ...
  getProvider?: ...
}

The separate interface feels more explicit, but CommonPlugin may be a more natural place given that a plugin can already provide multiple general-purpose capabilities.

**2. Is providerPriority necessary?**

The priority mechanism is included in this draft to provide deterministic ordering when multiple plugins register providers.

However, I'm not sure whether Zudoku should expose ordering as part of the public plugin API or whether provider order should instead follow plugin declaration order.

If ordering should simply follow:

plugins: [pluginA, pluginB]

then providerPriority can be removed.

**3. What should the provider wrap?**

The current implementation makes children represent the actual Zudoku application content, while the custom provider itself is rendered inside the existing Zudoku provider hierarchy.

In other words:

ZudokuProvider
└── existing Zudoku providers
    └── CustomProvider
        └── Zudoku application

This was intentional so custom providers can access Zudoku's React context if they need it.

I'd appreciate feedback on whether this is the appropriate placement or whether custom providers should wrap a larger portion of the Zudoku tree.

**4. Should getProvider be required?**

Currently the interface uses:

getProvider?: ...

because it follows the optional nature of the other plugin capabilities.

However, if ProviderPlugin is only identified through hasProvider, it may make more sense for getProvider to be required on ProviderPlugin.

fixes #1458

**Status**

This PR is intentionally opened as a draft to get feedback on the API/design before adding more extensive tests or refining the implementation.

The main goal at this stage is to validate whether the general approach of allowing plugins to register React Providers fits Zudoku's plugin architecture.


mosch:
That makes sense, happy to get it in when done

tusharmotwani1718:
hey @mosch the pr is ready to be reviewed, have tested by rendering providers without and with different priorites.

once reviewed, can update the docs too mentioning the use case and examples.

PR #2778, opened 2026-08-25
Add disableMcpAuthInstructions option for MCP servers

mosch:
## Summary

Adds a new configuration option `disableMcpAuthInstructions` that hides authentication instructions on MCP server endpoint cards. This is useful when credentials are handled through other means (proxy injection, a separate login flow, etc.) and should not be documented in the install snippets.

## Key Changes

- **New `resolveMcpAuth()` function** in `mcp-configs.ts` that decides what auth the card renders. When `disableAuthInstructions` is set it short-circuits to `authType: "none"` before inspecting the security data at all, so every auth type is dropped — not just API keys.

- **Updated `MCPEndpoint` component** to use `resolveMcpAuth()` instead of separate `getAuthType()` and `getAuthHeader()` calls, passing through the `disableAuthInstructions` prop.

- **Configuration plumbing** across the stack:
  - Added `disableMcpAuthInstructions` to `ApiOptionsSchema` (Zod validator)
  - Added to `BaseOasConfig` interface in `interfaces.ts`
  - Threaded through `OperationListItem` → `MCPEndpoint`
  - Added to Vite plugin config generation (per-API `options` and `defaults.apis`)

- **Documentation updates**:
  - New "Authentication instructions" section in the `mcp-servers.md` guide explaining the feature and how to use it
  - Updated API reference with the new option description
  - Updated `x-mcp-server` extension docs to mention the option

- **Test coverage** for `resolveMcpAuth()`: API key and OAuth resolution when enabled, both dropped when disabled, and the no-security case.

## Implementation Details

- When disabled, the card renders as if the server were unauthenticated: no credential header in the snippets, no "replace `YOUR_API_KEY`" step, and the full client list is available again (Claude Desktop, ChatGPT, and the one-click Cursor/VS Code install links come back, since the reason for hiding them was that they cannot carry a custom header).
- OAuth already renders identically to no auth in today's card — `getAuthHeader()` returns nothing for `oauth2`/`openIdConnect` and every client lists `oauth` alongside `none` in `supportedAuth` — so the option is a no-op for OAuth servers in practice. It is still handled behind the same switch so the option means what its name says, and any auth-specific rendering added later stays behind it.
- The option can be set per-API in the `options` object or globally under `defaults.apis`.

## Verification

- Full test suite green (1609 tests, 134 files); typecheck, biome, and oxfmt clean.
- Rendered the real card in the `cosmo-cargo` example, whose `/catalog/api-ai-cargo` MCP operation is `X-API-Key`-secured. Default: header note under the URL, `--header 'X-API-Key: YOUR_API_KEY'` in the Claude Code command, ChatGPT/Claude Desktop hidden. With the option on: none of that, all six clients listed. The example config was only touched for that check and left unchanged in the diff.

https://claude.ai/code/session_017YkGgm8yKmP74mohys1q6b


vazexqi:
Reviewed and Approved

PR #2780, opened 2026-08-25
refactor(theme): unify badge colors onto one themeable palette

### LINKED ISSUE #2760: NavigationBadge: hardcoded colors, impossible to override (contrast issues)
The [NavigationBadge](https://github.com/zuplo/zudoku/blob/c62ed70f948addbce43939fba8e8e7d2f2603a3f/packages/zudoku/src/lib/components/navigation/NavigationBadge.tsx#L1-L24) component defines its colors via hardcoded Tailwind classes in ColorMap and ColorMapInvert:

```js
export const ColorMap = {
  green: "bg-green-400 dark:bg-green-800",
  blue: "bg-sky-400 dark:bg-sky-800",
  yellow: "bg-yellow-400 dark:bg-yellow-800",
  red: "bg-red-400 dark:bg-red-800",
  purple: "bg-purple-400 dark:bg-purple-600",
  indigo: "bg-indigo-400 dark:bg-indigo-600",
  gray: "bg-gray-400 dark:bg-gray-600",
  outline: "border border-border rounded-md text-foreground",
};
```

Unlike the rest of the Zudoku design system, these colors don't go through CSS theme tokens (--color-* / custom variables), which means:

It's not possible to override them through the site's theme configuration.
Although the component accepts a className prop, it's only merged with the existing classes (via cn) rather than replacing them, so it can't easily be used to override the colors either without resorting to !important.

This causes real contrast issues depending on the theme/branding in use: for example, the PATCH badge has a measured text/background contrast ratio of 3.79, below the WCAG AA threshold of 4.5:1 recommended for small text.

Image

PATCH badge with insufficient contrast (3.79), measured via DevTools

## Expected behavior

Be able to override navigation badge colors (via site theme configuration, CSS variables, or a dedicated prop), to allow reaching sufficient contrast depending on the branding in use.

Ideally, have ColorMap / ColorMapInvert rely on CSS theme tokens rather than fixed Tailwind classes, in line with the rest of the design system's components.

## Context

packages/zudoku/src/lib/components/navigation/NavigationBadge.tsx

mosch:
Supersedes #2767 — same goal (make badge colors themeable, fixes #2760), but derives the treatments from one palette instead of relocating the hardcoded values into 14 variables.

## The problem

Badge colors lived in three disjoint hardcoded maps for the same seven semantic colors:

| System | Light | Dark | Used by |
|---|---|---|---|
| `ColorMap` (fill) | `{c}-400` | `-800` green/blue/yellow/red, `-600` purple/indigo/gray | filled nav badges |
| `ColorMapInvert` (text) | `-500` green, `-400` rest | `{c}-600` | inverted nav badges |
| `TextColorMap` | `{c}-600` | *(none)* | method labels in `Sidecar` / `OperationListItem` |

None could be overridden by a site, and they had drifted apart from each other.

## The change

One `--badge-*` token per color. Both treatments derive from it via `color-mix()` in `main.css`, the same way `--callout-*` already works. A site re-colors every badge surface by setting seven variables.

A treatment is produced by shading the token toward the dark end of the theme, and mixing can only darken — so the palette holds the light, saturated shade and both treatments derive downward from it:

```css
--badge-shade: var(--foreground);   /* .dark: var(--background) */
--badge-fill-mix: 100%;             /* .dark: 50%  */
--badge-text-mix: 50%;              /* .dark: 100% */
```

The two derived values are registered as `@theme inline` colors (`bg-badge-fill`, `text-badge-text`), so they behave like ordinary Tailwind color utilities and `cn()`/twMerge resolves them against other `text-*` classes rather than leaving both in the class list. `--badge-color` is set per element, and `inline` is what makes it resolve there rather than at `:root`.

The mix values are chosen by the hue that needs it most (always yellow), swept against both the default theme and `cosmo-cargo` — the mix shades toward the theme's own `--foreground`, so a site with a lighter foreground lands lighter.

## What this changes visually

Only **light-mode filled badges** render identically to `main`; the token is the Tailwind `-400` value verbatim and `--badge-fill-mix` is 100%. Everything else moves:

| Surface | Light | Dark |
|---|---|---|
| Filled nav badge | **unchanged** | **changed** — 50% token over `--background` instead of `-800`/`-600` |
| Inverted nav badge | **changed** — was `-400`/`-500`, now a 50% mix | **changed** — was `-600`, now the token |
| Method label | **changed** — was exactly `-600`, now a 50% mix | **changed** — previously had no dark variant at all |

A single mix can't reproduce Tailwind's `-600` ramp across hues, because those ramps aren't constant-hue — yellow drifts most (91.9° vs 75.8°, lime vs amber). What the mix targets instead is contrast: at 50% every hue clears WCAG AA as label text on both `--background` and `--sidebar`, and every hue beats what it replaced.

## Fixes that fall out of having one source of truth

- **OPTIONS** rendered indigo in the operation list but gray in the sidebar. Both are indigo now.
- **TRACE** had no entry in the method map, but the lookup carried a `noNonNullAssertion` ignore claiming it "is guaranteed to be defined". It isn't — a TRACE operation rendered a colorless badge. The lookup falls back to gray and the assertion is gone.
- **Method labels had no dark-mode variant at all** and rendered their light-mode shade on a dark background.
- **Dark fills were perceptually uneven**: purple and indigo sat at chroma 0.288/0.262 against ~0.12 for everything else, so PATCH glowed next to GET. All seven now sit near 0.10.
- **An inverted `outline` badge** emitted no border and near-white text on a near-white background. `outline` now ignores `invert`, which has no meaning without a fill.

## Contrast

Measured from the built page. Light labels (vs `--background` / vs `--sidebar`) on `cosmo-cargo`, the tighter of the two themes:

```
green  4.93 / 4.80    blue   5.62 / 5.47    yellow 4.58 / 4.46
red    6.58 / 6.41    purple 6.43 / 6.26    indigo 6.86 / 6.68    gray 6.23 / 6.06
```

Dark clears AA everywhere (labels 5.57–11.08, white-on-fill 4.95–7.44).

White-on-fill for light-mode *filled* badges is 1.55–3.08. That is unchanged from `main` — the light fill is the token verbatim — and is worth its own change rather than this one.

## Verification

- `pnpm -F zudoku typecheck`, `pnpm biome ci` on the changed files, `pnpm fmt:check` — clean. (The two `noDescendingSpecificity` warnings in `main.css` are pre-existing on `main`.)
- `pnpm vitest run` — 1609 passed / 134 files.
- Built `cosmo-cargo` and confirmed the utilities are emitted, with Lightning CSS auto-generating non-`color-mix` fallbacks.
- Read computed colors and contrast ratios out of the rendered page for all seven colors × both treatments × both themes.
- Screenshotted the sidebar on this branch and on `main`, light and dark.

## Docs

New **Badge Colors** section in `colors-theme.mdx`: the seven tokens and which method each drives, how to set them through `customCss`, and a table for the four derivation variables.

## How to test

1. `nx run cosmo-cargo:dev`, open any OpenAPI page, toggle light/dark.
2. Add `--badge-green: red;` to your site's CSS — GET badges and GET method labels both follow it, in both themes and both treatments.

## Note on #2767

That PR found a real dark-mode bug: moving the invert colors into variant-less classes lets `dark:text-zinc-50` survive `twMerge` and win in dark mode. Its stated reason is off — twMerge *does* classify `text-badge-*` as a text color (it removes `text-background`); the cause is that a variant-less class can't displace a `dark:`-prefixed one. This branch avoids the situation entirely: the base class no longer carries an unconditional `text-background dark:text-zinc-50`, so there's nothing to displace.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GVoNhp8NJkzQMrTxqVZzEk

---
_Generated by [Claude Code](https://claude.ai/code/session_01GVoNhp8NJkzQMrTxqVZzEk)_


mosch:
## Review: this changes colors as well as making them configurable

The token/`color-mix()` structure is clean and mirrors `--callout-*`. `ColorMap` / `ColorMapInvert` / `TextColorMap` were never re-exported from `packages/zudoku/src/lib/components/index.ts`, so removing them isn't a public API break.

But it's worth being explicit that only **light-mode filled badges** render identically to `main`. Everything else moves:

| Surface | Light | Dark |
|---|---|---|
| Filled nav badge | **unchanged** — tokens are the Tailwind `-400` oklch values verbatim and `--badge-fill-mix: 100%` | **changed** — 47% token over `--background` instead of `-800`/`-600` |
| Inverted nav badge | **changed** — was `-400`/`-500`, now ≈`-500`/`-600` | **changed** — was `-600`, now the `-400` token |
| Method label (`Sidecar`, `OperationListItem`) | **changed** — was exactly `-600`, now ≈`-500` | **changed** — previously had no dark variant at all |

Plus the three semantic changes already in the description (OPTIONS gray → indigo in the sidebar, TRACE colorless → gray, `outline` ignoring `invert`).

### 1. "Light-mode label text on the same `-600` ramp" isn't quite right

A single 78% mix can't reproduce Tailwind's `-600` across hues — matching green needs ~59%, red ~77%. Contrast on white, computed from the derived values:

```
green   4.94:1 → 3.03:1     blue   5.85:1 → 3.65:1
gray    7.56:1 → 4.25:1     red    4.76:1 → 4.61:1
```

These are ~12px mono labels in the sidecar and operation list, and green/blue/gray all drop from passing AA to failing it. Yellow also shifts hue 75.8° → 91.8° (amber → lime).

Cheap fix: drop `--badge-text-mix` to ~60–65%, or give the low-contrast hues a per-token override.

### 2. The dark fills move more than "slightly brighter" suggests

They are uniform, but the uniformity comes from flattening chroma: purple 0.288 → 0.098, red 0.177 → 0.090, and purple also drops in lightness (L 0.558 → 0.410). Fill-vs-white-text contrast improves, but the seven colors are now separated almost entirely by hue. Worth a side-by-side screenshot check before merging — that one's a taste call.

### 3. Docs gap

`docs/pages/docs/customization/colors-theme.mdx` still lists only the shadcn core variables, so the new `--badge-*` tokens aren't documented anywhere. Given that configurability is the point of the PR (#2760), that seems worth adding there and/or in `docs/pages/docs/components/badge.mdx`.

---

mosch:
Thanks — the contrast point was right and I'd missed it. I'd checked lightness and chroma bands but never computed contrast ratios. Fixed in 5448a31.

### 1. Contrast — fixed, both mixes now 50%

I reproduced the measurements and my baselines for `main` differ from yours on two hues (green-600 on white I get 3.22 not 4.94, blue-600 4.02 not 5.85; gray 7.56 and red 4.76 match). But that doesn't change the conclusion — it makes it slightly worse, since green and yellow were already under AA on `main`.

Rather than tune to a shade number, I swept for the binding hue (always yellow) against both the default theme and `cosmo-cargo`, which matters because the mix shades toward the theme's own `--foreground` — `cosmo-cargo` uses `#262626` where the default is `oklch(0.141 …)`, so it lands lighter. 55% would have cleared AA on the default theme but not on our own example.

At **50%**, every hue clears AA on both surfaces and every hue beats what it replaced, in both treatments. Rendered from the built page:

```
              label vs bg   label vs sidebar        (light)
green   4.93 / 4.80     blue   5.62 / 5.47
yellow  4.58 / 4.46     red    6.58 / 6.41
purple  6.43 / 6.26     indigo 6.86 / 6.68     gray 6.23 / 6.06
```

Worst cell is the yellow nav badge at 4.46 on `cosmo-cargo` — a hair under, up from 1.51 on `main`. Dark mode clears AA everywhere (labels 5.57–11.08, fills 4.95–7.44).

You're right that a single mix can't reproduce the `-600` ramp — I've dropped that claim from the description. The yellow hue drift (91.9° vs 75.8°) is the same root cause: Tailwind's ramps aren't constant-hue, so any single-token derivation will drift. Per-hue text overrides would fix it exactly, at the cost of going back toward two tokens per color.

### 2. Dark fills — raised to 50%, keeps the saturation

Good catch that 47% was flattening chroma harder than needed. 50% holds chroma at ~0.104 against the ~0.10–0.12 the old `-800` fills carried, so the colors stay separable by saturation and not hue alone, while white-on-fill still clears AA (5.83:1 worst hue on the default theme, 4.95:1 on `cosmo-cargo`). Purple and indigo still lose their outlier status — that part was the point.

### 3. Docs — added

New **Badge Colors** section in `colors-theme.mdx`: the seven tokens and which method each drives, how to set them via `customCss`, why a token should be the light saturated shade, and a table for the four derivation variables with a note to lower the mixes if you override with lighter colors.

### One thing I did not change

White-on-fill for light-mode *filled* badges is 1.55–3.08. That's unchanged from `main` — the light fill is the token verbatim, so those badges are byte-identical — and fixing it means changing the one surface that currently doesn't move. Happy to do it as a follow-up if you want it.

---

PR #2797, opened 2026-09-01
docs: filter links excluded from Zuplo docs

ntotten:
## Summary

- hide the Vercel-only output note when Zudoku content is rendered in Zuplo mode
- use the canonical MCP guide path so the Zuplo sync can redirect it to the Zuplo-specific guide

The immediate downstream sync fix is zuplo/docs#1318.

## Verification

- pnpm run fmt:check
- pnpm check:other
- production docs build
- built-site link check: 106 links, 0 errors


vazexqi (commented):
Reviewed and approved

PR #2806, opened 2026-09-16
Prevent React from outlining suspended content in SSR

mosch:
## Summary

Fixes an issue where React's streaming SSR was outlining large suspended Suspense boundaries into hidden segments swapped in by inline scripts, making the content invisible to clients without JavaScript (AI crawlers, curl, reader mode).

## Changes

- **`entry.server.tsx`**: Set `progressiveChunkSize: Number.MAX_SAFE_INTEGER` when calling `renderToReadableStream()`. Since the stream is only read after `allReady` completes (no progressive streaming), this prevents React from outlining completed boundaries larger than the default chunk size (12.8 kB) into hidden segments.

- **`entry.server.test.tsx`**: Added test coverage for the fix:
  - Added `DeferredPage` component and `/deferred` route that renders a Suspense boundary with deferred content
  - Added test case `"inlines large suspended content instead of deferring it to a script"` that verifies:
    - The deferred content is rendered inline in the HTML
    - The loading fallback is not present in the final output
    - React's hidden segment markers (`<div hidden id="S:` and `<template id="B:`) are not present

## Implementation Details

The root cause was React's default behavior of outlining large completed Suspense boundaries into hidden segments to reduce initial HTML size. However, in Zudoku's SSR model where the entire stream is buffered before being sent to the client, this optimization is counterproductive and breaks accessibility for non-JS clients.

By setting `progressiveChunkSize` to `Number.MAX_SAFE_INTEGER`, we effectively disable this outlining behavior, ensuring all content is inlined in the HTML where it's visible to all clients.

https://claude.ai/code/session_016nuUGHbWNabY7qeh7MbZCJ


lcampos (commented):
Reviewed and approved

PR #2808, opened 2026-09-16
fix: keep the SSR entry out of the client dependency scan

ntotten:
## What's broken

Every page of a zudoku dev server logs, repeatedly:

```
Uncaught SyntaxError: The requested module '/docs/node_modules/style-to-js/cjs/index.js?v=0a1d1e3a' does not provide an export named 'default'
Uncaught SyntaxError: The requested module '/docs/node_modules/debug/src/...' does not provide an export named 'default'
```

Reported against 0.87.0 from the zuplo/docs site, on every page type. Production builds are unaffected — rolldown does the CJS/ESM interop at build time.

## Root cause

Vite's dependency scan was **crashing outright**, so pre-bundling was skipped entirely:

```
(!) Failed to run dependency scan. Skipping dependency pre-bundling. Error: The following
dependencies are imported but could not be resolved:
  virtual:zudoku-markdown-files (imported by .../zudoku/src/app/entry.server.tsx)
```

The chain:

1. `optimizeDeps.entries` was `src/{app,lib}/**` — broad enough to pull `src/app/entry.server.tsx` into the **client** scan.
2. That file imports `virtual:zudoku-markdown-files`, provided by a plugin gated on `applyToEnvironment(env) => env.name === "ssr"` (`plugin-markdown-export.ts:223`). It cannot resolve in the client environment.
3. Vite aborts the entire scan on one unresolvable import, so nothing was pre-bundled beyond the explicit `include` list.

Why this only bites installed sites: `zudoku` is in `optimizeDeps.exclude`, so its source is served unbundled and its bare imports must be pre-bundled by name. Vite's runtime-discovery fallback has a hard guard — `importer && isInNodeModules(importer)` — so for importers under `node_modules` it skips `registerMissingImport` and instead serves the raw file with `injectQuery(resolved, 'v=' + browserHash)`. That is exactly the `?v=0a1d1e3a` in the error. In this monorepo zudoku's realpath sits *outside* `node_modules`, so runtime discovery silently recovered — at the cost of 504 "Outdated Optimize Dep" churn and full page reloads. Same crash, invisible symptom.

## The fix

Narrow the scan entries to `.ts`/`.tsx` and exclude the SSR entry (plus test files, so local dev scans the same file set an installed site does).

A broad superset is kept deliberately rather than seeding only the client entry: in an installed site the scan is the *only* discovery path, so anything it misses is unrecoverable at runtime.

## Verification

Built a faithful consumer repro — `npm pack`'d zudoku unpacked as a real directory under `node_modules`, outside the pnpm workspace globs — plus a crawler that walks the served module graph and flags any dependency served as raw CommonJS.

| | before | after |
|---|---|---|
| scan failures | 1 | 0 |
| deps pre-bundled | 27 | 261 |
| modules loaded per page | 6180 | 441 |
| raw-CJS deps served | **8** | **0** |

The 8 were `style-to-js`, `debug`, `hoist-non-react-statics`, `parse5`, `@braintree/sanitize-url`, `cytoscape`, `cytoscape-fcose`, `cytoscape-cose-bilkent` — the two reported errors plus six more. `debug` arrives via mermaid → cytoscape.

Also checked:

- Dev console clean on article / MDX / OpenAPI operation / catalog pages, in both the consumer repro and `cosmo-cargo`.
- `nx run cosmo-cargo:build` and a production preview serve clean.
- `pnpm check`, `pnpm -F zudoku typecheck`, `pnpm test` (1858 passed), `pnpm size` (121.59 kB / 215 kB limit).

## Reviewer notes

- The negative glob patterns rely on Vite 8's `globEntries`, which routes patterns to two `tinyglobby` calls split on whether the pattern string contains `node_modules`. All the patterns here derive from the same root, so they always land in the same bucket and the negations apply — verified in both the monorepo and installed layouts (identical 417 entries).
- The new test is a real regression guard: it fails against the old single-glob behaviour.
- **Not a bug, checked while here:** a consuming site's own `optimizeDeps.include` *is* merged and honored (verified with `react-dom/server`). An earlier attempt to work around this with `include: ["style-to-js"]` in the site's `vite.config.ts` failed only because `style-to-js` is a transitive dep of zudoku, not of the site, so it isn't resolvable from the site root — Vite logs `Failed to resolve dependency: style-to-js, present in client 'optimizeDeps.include'`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ntotten:
### Re: the coverage report showing −2.74%

Measured locally on this branch vs. its parent commit — the drop is a measurement-scope artifact, not a reduction in testing:

| | parent | this PR | delta |
| --- | ---: | ---: | ---: |
| Lines **covered** | 5960 | 6000 | **+40** |
| Lines **total** | 8408 | 8805 | **+397** |
| Lines % | 70.88% | 68.14% | −2.74% |

Covered lines went **up** by 40. The percentage fell because the denominator grew by 397.

Cause: `config.test.ts` previously imported only `loadZudokuConfig` from `../config/loader.js`. The new test imports from `./config.js`, which pulls `config.ts` and its transitive import graph (`build.js`, `plugin.js`, `protected/annotator.js`, `zuplo.js`, …) into v8's instrumented set for the first time. That vite-plugin code was always untested — it just wasn't being measured before.

So nothing became less tested; some previously-invisible untested code is now visible in the metric. I've deliberately left this alone rather than narrowing the test to avoid importing the module under test, since that would make the test worse to make the number look better.

All checks pass, and Copilot's review generated no comments.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

mosch (approved):

